PR #13, opened 2019-10-17
Pal 188 checkstyle

m09526:


code-style/checkstyle.xml
```diff
@@ -0,0 +1,142 @@
...
+        <module name="EmptyStatement"/>
+        <module name="EqualsHashCode"/>
+        <module name="EqualsAvoidNull"/>
+        <module name="IllegalInstantiation"/>
+        <module name="MissingSwitchDefault"/>
+        <module name="DefaultComesLast"/>
+        <module name="SimplifyBooleanExpression"/>
+        <module name="SimplifyBooleanReturn"/>
+        <module name="ConstantName"/>
+        <module name="FinalClass"/>
```
dev-958: not a fan of the finalClass rule, it renders the code inflexible.

dev-958 (changes requested):
Only one comment but also noticed a problem with the package naming in the policy service:
...cy-service/src/main/java/uk.gov.gchq.palisade.service.policy/request/AddCacheRequest.java 
now would probably be a good moment to fix that.

dev-958 (approved):

PR #27, opened 2019-11-05
Pal 238 Added -exec jar to discovery service

dev930018:
This is expected by the services-launcher


dev-958 (approved):

dev930018:
While git won't show it so easily, the pom originally had DOS line endings (\r\n) - this has been amended

m09526 (approved):

PR #32, opened 2019-11-13
Pal 202 data service unit tests

nw1984:


data-service/src/test/java/uk/gov/gchq/palisade/service/data/repository/SimpleCacheServiceTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class SimpleCacheServiceTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SimpleCacheServiceTest.class);
+
+    private SimpleCacheService cacheService = new SimpleCacheService();
+    private ServiceState serviceState = new ServiceState();
+    private BackingStore store = new HashMapBackingStore();
+    private Duration maxLocalTTL = Duration.of(5, ChronoUnit.MINUTES);
+
+    @Test(expected = NoConfigException.class)
+    public void applyConfigFromReturnsErrorTest() {
```
m09526: This test is no longer required. The applyFrom...() methods are being removed as they are part of the old config injection code.

data-service/src/test/java/uk/gov/gchq/palisade/service/data/service/SimpleDataServiceTest.java
```diff
@@ -0,0 +1,297 @@
...
+@RunWith(JUnit4.class)
+public class SimpleDataServiceTest {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SimpleDataServiceTest.class);
+
+    private AuditService auditService;
+    private SimpleCacheService cacheService;
+    private PalisadeService palisadeService;
+    private SimpleDataService simpleDataService;
+    private DataReader dataReader;
+    private ServiceState serviceState = new ServiceState();
```
m09526: Remove servicestate. It's legacy.

data-service/src/test/java/uk/gov/gchq/palisade/service/data/service/SimpleDataServiceTest.java
```diff
@@ -0,0 +1,297 @@
...
+            assertEquals(leafResource, generatedLeafResource);
+        }
+    }
+
+    private void setupCacheService() {
+        final BackingStore store = Mockito.mock(BackingStore.class);
+        serviceState.put("cache.svc.store", null);
+        serviceState.put("cache.svc.max.ttl", Duration.of(5, ChronoUnit.MINUTES).toString());
+        cacheService = new SimpleCacheService();
+        cacheService.recordCurrentConfigTo(serviceState);
```
m09526: Remove this line.

m09526 (changes requested):

palisade-service/mvn_dependency_tree.txt
```diff
@@ -136,13 +136,13 @@ io.etcd/jetcd-launcher/version = 0.3.0
 
 io.etcd/jetcd-resolver/version = 0.3.0
 
-io.fabric8/kubernetes-client/version = 4.3.0
+io.fabric8/kubernetes-client/version = 4.2.0
```
m09526: Please fix downgraded dependency.

resource-service/mvn_dependency_tree.txt
```diff
@@ -62,7 +62,7 @@ com.google.inject/guice/version = 4.1.0
 
 com.google.inject.extensions/guice-servlet/version = 4.0
 
-com.google.protobuf/protobuf-java/version = 3.5.1
+com.google.protobuf/protobuf-java/version = 2.5.0
```
m09526: Downgraded dependency.

m09526 (changes requested):
Two dependency downgrades to fix then approve.

palisade-service/mvn_dependency_tree.txt
nw1984: Running mvn dependency:tree (-X) shows the following:
```
[INFO] +- io.fabric8:kubernetes-client:jar:4.2.0:compile
```
This is a dependency from the top level i.e a dependency introduced by the parent:
```
    <parent>
        <artifactId>spring-boot-starter-parent</artifactId>
        <groupId>org.springframework.boot</groupId>
        <version>2.1.8.RELEASE</version>
    </parent>
```

See:
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/pom.xml

It could be that the mvn dependency chain has been updated to fix a comptability issue identified in the later build.

resource-service/mvn_dependency_tree.txt
nw1984: Dependency tree shows:
protobuf at 3.5.1 for spring-boot-starter-parent but
for hadoop-common the version is +- org.apache.hadoop:hadoop-common:jar:3.2.0:compile ==> 2.5.0
We cannot fix the version used by Hadoop.

m09526: The data service should no longer need the Hadoop dependency once the Hadoop reader has been moved from it to run separately. Is the data reader still a part of the data service at the moment? If so, we can leave this and it will disappear once the separation work is carried out.

palisade-service/mvn_dependency_tree.txt
m09526: Probably just that they haven't updated that version in the time that version of Spring was released. We'll leave this for now, it should upgrade as part of moving to new Spring releases.

resource-service/mvn_dependency_tree.txt
nw1984: Hi @m09526 - the resourceService (this dependency error is for the resource service) contains the HadoopResourceService which makes use of Configuration for instance which is part of the hadoop-common library. Does this seem incorrect???

PR #34, opened 2019-11-15
Pal 154 create template helm chart

dev-958:
Creates ingress and templates for multi-tenant installations with routing. More documentation on integration of the Palisade service.


charts/traefik/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+container:
+  args:
+    - --api.insecure
```
m09526, line 17: Do we need to enable the insecure API access?

m09526 (commented):
One clarification point, then approved.

charts/traefik/values.yaml
dev-958: Temporarily yes, there is a later ticket to secure this endpoint with SSL but it would remain insecure for local deployments (localhost) for developers testing palisade.

m09526 (dismissed):

PR #40, opened 2019-11-21
PAL-258 Logging improvements to palisade service

dev930018:
As this may be the first of many similar, it is worth ensuring the initial set of changes to the palisade service are complete

The ticket suggests:
> 1) Log at INFO level for receipt of a message
> 2) Log at INFO level for a message sent to another service
> 3) Log at DEBUG for individual function calls with more metadata
> 4) In the event of ERRORs include the METAdata of what was being handled

If this is the full set of cases for logging, we should only see messages on INFO if services are communicating with one another


palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/service/AuditServiceTest.java
```diff
@@ -0,0 +1,93 @@
...
+    private AuditService auditService;
+
+    @Mock
+    AuditClient auditClient;
+    @Mock
+    Executor executor;
+
+    @Before
+    public void setUp() {
+        logger = (Logger) LoggerFactory.getLogger(AuditService.class);
```
m09526, line 55: Unecessary cast

m09526 (changes requested):

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/service/AuditServiceTest.java
dev930018: The Logback Loggers allow us to add appenders and listen to logging statements, but not in slf4j

This is more verbosely ch.qos.logback.Logger logger = org.slf4j.LoggerFactory.getLogger(~.class);

dev930018: The name is a bit misleading, but the cast is needed

m09526 (dismissed):

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/config/ApplicationConfiguration.java
```diff
@@ -155,6 +161,7 @@ public ObjectMapper objectMapper() {
     @Bean(name = "eureka-client")
     @ConditionalOnProperty(prefix = "eureka.client", name = "enabled")
     public ServiceInstanceRestController eurekaClient() {
+        LOGGER.debug("Instantiated eurekaClient");
```
dev-958: INFO please, we would always want to see this at startup

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/config/ApplicationConfiguration.java
```diff
@@ -164,7 +171,7 @@ public Executor getAsyncExecutor() {
         return Optional.of(new ThreadPoolTaskExecutor()).stream().peek(ex -> {
             ex.setThreadNamePrefix("AppThreadPool-");
             ex.setCorePoolSize(6);
-            LOGGER.info("Starting ThreadPoolTaskExecutor with core = [{}] max = [{}]", ex.getCorePoolSize(), ex.getMaxPoolSize());
+            LOGGER.debug("Starting ThreadPoolTaskExecutor with core = [{}] max = [{}]", ex.getCorePoolSize(), ex.getMaxPoolSize());
```
dev-958: ditto

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/ResultAggregationService.java
```diff
@@ -119,7 +123,7 @@ private void auditRequestReceivedException(final RegisterDataRequest request, fi
                         .withContext(request.getContext())
                         .withException(ex)
                         .withServiceClass(serviceClass);
-        LOGGER.debug("Error handling: " + ex.getMessage());
+        LOGGER.debug("Auditing request exception: \n\t{}\n\t{}", request, ex);
```
dev-958: Here we are logging an exception event, we will only be here if an exception has occurred so we will certainly want to see that meta data whatever the level so ERROR level is applicable.

dev-958 (changes requested):
see inline comments

m09526 (approved):

dev-958 (approved):

PR #41, opened 2019-11-22
Pal 181 create helm charts palisade

dev-958:
Includes some documentation


nw1984 (approved):
Nice work

charts/palisade-service/templates/deployment.yaml
```diff
@@ -0,0 +1,75 @@
...
+                          path: /actuator/health
+                          port: 8084
+                      initialDelaySeconds: 10
+                      failureThreshold: 5
+                      periodSeconds: 20
+                      timeoutSeconds: 5
+                  readinessProbe:
+                      httpGet:
+                          path: /actuator/health
+                          port: 8084
```
m09526, line 57: Palisade service used to be on 8080, is there a need to move it to 8084?

m09526 (changes requested):

charts/palisade-service/templates/deployment.yaml
dev-958: Pods are exposed to the cluster via the service on port 80 so the internal port is immaterial. Also convention places dashboards on 8080, as is the traefik dashboard (external ip) so wanted to avoid confusion.

m09526 (approved):

PR #42, opened 2019-11-25
PAL-275 move hadoop resource service

nw1984:


resource-service/pom.xml
```diff
@@ -35,15 +35,16 @@
...
     <dependencies>
-        <dependency>
-            <groupId>com.netflix.eureka</groupId>
-            <artifactId>eureka-client</artifactId>
-            <scope>compile</scope>
-        </dependency>
         <dependency>
             <groupId>io.etcd</groupId>
             <artifactId>jetcd-core</artifactId>
             <version>0.3.0</version>
```
m09526: Extract version

services-launcher/.log
```diff
@@ -0,0 +1 @@
+Error: Invalid or corrupt jarfile target
```
m09526: This file shouldn't be here.

m09526 (changes requested):

m09526 (dismissed):

resource-service/pom.xml
```diff
@@ -32,18 +32,20 @@
         <hadoop.version>3.2.0</hadoop.version>
         <fabric8.version>4.3.0</fabric8.version>
         <protobuf.version>3.5.1</protobuf.version>
+        <jetcd.core>0.3.0</jetcd.core>
     </properties>
 
     <dependencies>
-        <dependency>
```
dev-958: Removing this is likely to break ServiceInstanceRestController. Even though it doesn't call any Feign clients.

dev-958 (changes requested):

PR #55, opened 2019-12-12
PAL-289 Create helm charts for data service

dev930018:
Added data service charts as copy of audit service charts with autoscaling


dev930018:
By mistake, this also includes PAL-295 user service helm chart fixes

PR #60, opened 2019-12-17
Pal 289 data service helm charts

dev930018:


dev930018:
Further required for this ticket:
- Ingress on /data
- Mount point for the associated data

dev-958 (changes requested):
As it stands the data service does not come up because you have not defined a PersistentVolume to bind with the claim. This needs to be added and parameterised from the top level. Also allow for there to be multiple locations (an array). The volumes that are mounted should be written to output on install and update the README accordingly.

dev-958 (approved):

dev-958 (changes requested):
PersistentVolume is not mounting

nw1984 (approved):

PR #74, opened 2020-02-07
Pal 389 user service springboot cacheable

developer01189998819991197253:
Deleted old cache mechanism, removed backingStores
Added Caffeine cache
Updated Copyright
Added CachedUserService for @Cacheable method
Removed CompletableFuture from userService 
Removed cache tests and have added them to Integration tests
Added UserControllerTest


pd104923 (dismissed):

dev-958 (changes requested):
Needs some redesign to the structure, spoke to Chris.

dev-958 (dismissed):

user-service/src/main/resources/application.yaml
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 #
 spring:
+  cache:
+    caffeine:
+      spec: expireAfterWrite=60s, maximumSize=500
```
ac74475: If this is the default config then doesn't this mean that if you have more than 60 seconds between configuring the example and running it, then you will hit user not found exceptions as we are using the null-user-service

ac74475: If my above assumption is correct then I suggest making the ttl 1hour

ac74475 (changes requested):
Why do we have an empty users.txt file?
Also there is no need to update the copyright year on each file unless you have modified the majority of the file.

user-service/src/main/resources/application.yaml
developer01189998819991197253: agreed to set ttl to 60m

ac74475 (approved):

PR #77, opened 2020-02-13
Pal 514 refactor launcher application yaml

pd104923:
Changed the launch process for the services within the services-launcher. This has been done due to the Spring and Eureka dependencies being taken from the example-library jar rather than the service exec jar using the current process.
The change for resolving this has only been added to the services that require the use of the example-library dependency jar (audit, data, palisade, policy, user).

The command for launching a service has been drastically reduced in length and now looks like this:
`java -Dloader.path=../palisade-examples/example-library/target/ -Dspring.profiles.active=jvm -jar palisade-service/target/palisade-service-0.4.0-SNAPSHOT-exec.jar`

### DIFF
--- a/audit-service/pom.xml
+++ b/audit-service/pom.xml
@@ -226,6 +226,9 @@
             <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
+                <configuration>
+                    <layout>ZIP</layout>
+                </configuration>
                 <executions>
                     <execution>
                         <id>repackage</id>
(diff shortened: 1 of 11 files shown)


ac74475 (changes requested):
We should add the ZIP fix to the resource service as well incase that service requires the example library at some point in the future

ac74475 (approved):

pd104923:
> We should add the ZIP fix to the resource service as well incase that service requires the example library at some point in the future

The configuration has been added to the resource-service pom file as well.

PR #97, opened 2020-03-11
PAL-561 set eureka to prefer hostname over ip

dev930018:


developer01189998819991197253 (approved):

dev930018:
McAfee is a nightmare...
10.1.11.xx and 192.168.1.xx seem to be blocked, but localhost is not
This change should declare each service to eureka under localhost rather than the equivalent IP

dev-958 (approved):

ac74475 (approved):

PR #101, opened 2020-03-17
PAL-546 updated services for example and caching changes

dev930018:


developer01189998819991197253 (dismissed):

services-manager/src/main/resources/application-debug.yaml
```diff
@@ -13,6 +13,10 @@
 # limitations under the License.
 #
 
+eureka:
```
ac74475: I would have thought this should only be in the eureka profile and not in the debug profile. I assume we will want to use debug profile with both JVM and K8 deployments

dev930018: The logging profile can do two things:
* Change what level is used on startup
* Ping all available /loggers endpoints to change level at runtime

This second feature requires eureka.
In the future, there may be a k8s oir jvm 'mode' switch seperately that could also include helm install commands etc. but regardless, the debug profile switch should be expecting eureka to be running in jvm mode

dev930018: I'll pull it out, because there are some other changes required to the scripts

dev930018:
The problem with the hr-data-generator having logging (last week?) was not appearing because there was no hr-data-generator in the example-library

services-manager/src/main/resources/application-eureka.yaml
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Crown Copyright
```
ac74475:
```suggestion
# Copyright 2020 Crown Copyright

```

services-manager/src/main/resources/application-services.yaml
```diff
@@ -20,23 +20,21 @@
...
-  client:
-    enabled: true
-
 manager:
   # Defaults used internally to the yaml ONLY - these are not interpreted by any java application, just to minimise duplication of long filepaths etc.
   default:
     # Library includes relative to manager.root above (e.g. ../palisade-examples => palisade-services/../palisade-examples)
     # While the example library happens to be the same version as services, it should be declared here as a static name
-    library: ../palisade-examples/example-library/target/
+    example-library: ../Palisade-examples/example-library/target/example-library-0.4.0-SNAPSHOT.jar
```
ac74475: any way to parameterise the versions here?

ac74475 (changes requested):

services-manager/src/main/resources/application-services.yaml
dev930018: See comment just above, yes but there is no guarantee one version matches another.
It could be wildcarded for now?

developer01189998819991197253 (approved):

m09526 (approved):

ac74475 (approved):

PR #109, opened 2020-03-25
Pal 394 use cache warming for example data

pd104923:
Will require a Palisade-common merge before building successfully.
Added postConstruct method to PolicyController and UserController.
Created StdUser Configuration and CacheWarmerFactory.


policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/request/Policy.java
```diff
@@ -1,225 +0,0 @@
-/*
```
developer01189998819991197253, line 1: Is there a reason this file is no longer needed?

services-manager/src/main/resources/application-example.yaml
```diff
@@ -0,0 +1,85 @@
+# Copyright 2020 Crown Copyright
```
developer01189998819991197253, line 1: Is there a reason to have example, services and jvm profiles? can they not be merged?

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
```diff
@@ -0,0 +1,116 @@
...
+import static java.util.Objects.requireNonNull;
+
+@ConfigurationProperties
+public class StdUserCacheWarmerFactory implements UserCacheWarmerFactory {
+
+    private String userId;
+    private Set<String> auths;
+    private Set<String> roles;
+
+    public StdUserCacheWarmerFactory() {
```
developer01189998819991197253, line 43: JavaDoc

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
```diff
@@ -0,0 +1,116 @@
...
+        this.auths = auths;
+        this.roles = roles;
+    }
+
+    public String getUserId() {
+        return userId;
+    }
+
+    public void setUserId(final String userId) {
+        this.userId = userId;
```
developer01189998819991197253, line 70: setter methods need requireNonNull

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+
+@ConfigurationProperties(prefix = "population")
+public class StdUserConfiguration implements UserConfiguration {
+
+    private List<StdUserCacheWarmerFactory> users = new ArrayList<>();
+
+    public StdUserConfiguration() {
+    }
+
+    public StdUserConfiguration(final List<StdUserCacheWarmerFactory> users) {
```
developer01189998819991197253, line 50: JavaDoc

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+        this.users = users;
+    }
+
+    @Override
+    public List<StdUserCacheWarmerFactory> getUsers() {
+        return users;
+    }
+
+    public void setUsers(final List<StdUserCacheWarmerFactory> users) {
```
developer01189998819991197253, line 61: setter method needs requireNonNull

developer01189998819991197253 (changes requested):

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/request/Policy.java
ac74475: It has now been put in common, which isn't ideal, but we couldn't work out a way around it yesterday

services-manager/src/main/resources/application-example.yaml
ac74475: I like that they are separate profiles.
Example - adds in the example jars so those beans can be used and example cache warming used
services - starts up the palisade services
discovery - starts up the eureka services
So unless you can create a profile jvm_example that just calls to the other profiles, I would leave this as is.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
pd104923: Used code generator template to update this

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
pd104923: Used code generator template to update this

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
pd104923: Added

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
pd104923: Added

services-manager/src/main/resources/application-example.yaml
pd104923: Leaving this as is for now.

policy-service/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=uk.gov.gchq.palisade.example.config.ApplicationConfiguration
```
ac74475: shouldn't this be the services application config rather than the example one?

services-manager/src/main/resources/application-example.yaml
```diff
@@ -0,0 +1,85 @@
...
+# All fields are *technically* optional but some (classpath, launcher, main) are probably required to start the spawned JVM process
+# Extensive use is made of the eureka discovery client to discover services and endpoints - subsequently the "eureka" profile is enabled by default for all services
+# Extract the project version (x.y.z-REVISION etc.)
+app:
+  version: '@project.version@'
+    # While the library happens to be the same version as services, it should be declared here as a static name
+library:
+  version: ${app.version}
+
+eureka:
```
ac74475: I have concerns around having the eureka stuff set in this example yaml. If we keep it with eureka then it needs to be renamed to JVM_example. However I think it is better to have the eureka stuff being in a separate spring profile, such that if you are running the JVM example then you need both the eureka stuff and the example. But if you are running the K8's example then you only need the example profile as k8 already has a discovery service.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
ac74475: I think you need to delete the existing setters and getters and regenerate. Certainly mine generate differently to these. You might also want to add mutator methods, given you currently have your setters return the StdUserCacheWarmerFactory object which setters shouldn't.

user-service/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=uk.gov.gchq.palisade.example.config.ApplicationConfiguration
```
ac74475: This should be the application configuration in services rather than the example one

ac74475 (changes requested):

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
pd104923: I think these were created using Adam's original templates. I'll generate them with the new templates and add some mutator methods

policy-service/src/main/resources/META-INF/spring.factories
pd104923: The services config already has the @EnableAutoConfiguration annotation applied to it so changing this to the service application config creates a circular import issue when starting up the service.
This is how the example config beans are loaded into the service.

user-service/src/main/resources/META-INF/spring.factories
pd104923: The services config already has the @EnableAutoConfiguration annotation applied to it so changing this to the service application config creates a circular import issue when starting up the service.
This is how the example config beans are loaded into the service.

services-manager/src/main/resources/application-example.yaml
pd104923: I shall work on taking the eureka stuff out of this profile so that this can work with the k8s example in the future.

policy-service/src/main/resources/META-INF/spring.factories
pd104923: Moved spring.factories to example-library.

user-service/src/main/resources/META-INF/spring.factories
pd104923: Moved spring.factories to example-library.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
```diff
@@ -0,0 +1,262 @@
...
+        URI normalised = FileUtil.convertToFileURI(resource);
+        String resource = normalised.toString();
+        if (resource.endsWith(".avro")) {
+            return new FileResource().id(resource).type(type).serialisedFormat("avro").parent(getParent(resource));
+        } else {
+            return new DirectoryResource().id(resource).parent(getParent(resource));
+        }
+    }
+
+    public ParentResource getParent(final String fileURL) {
```
ac74475:
```suggestion
    private ParentResource getParent(final String fileURL) {
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
```diff
@@ -0,0 +1,262 @@
...
+                //else recurse up a level
+                return new DirectoryResource().id(parent.toUri().toString()).parent(getParent(parent.toUri().toString()));
+            }
+        } else {
+            //if this is another scheme then there is no definable parent
+            return new SystemResource().id("");
+        }
+    }
+
+    public boolean isDirectoryRoot(final Path path) {
```
ac74475:
```suggestion
    private boolean isDirectoryRoot(final Path path) {
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
pd104923: Updated

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
pd104923: Updated

ac74475 (dismissed):

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
```diff
@@ -0,0 +1,262 @@
...
+    private String owner;
+    private Map<String, String> resourceRules;
+    private Map<String, String> recordRules;
+
+    public StdPolicyCacheWarmerFactory() {
+    }
+
+    public StdPolicyCacheWarmerFactory(final String type, final String resource, final String owner,
+                                       final Map<String, String> resourceRules, final Map<String, String> recordRules) {
+        this.type = type;
```
developer01189998819991197253, line 66: Use the setters and getters created below

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Constructor with 2 arguments for a standard implementation
+     * of the {@link PolicyConfiguration} interface
+     *
+     * @param policies  a {@link List} of objects implementing the {@link uk.gov.gchq.palisade.service.PolicyCacheWarmerFactory} interface
+     * @param users  a {@link List} of objects implementing the {@link uk.gov.gchq.palisade.service.UserCacheWarmerFactory} interface
+     */
+    public StdPolicyConfiguration(final List<StdPolicyCacheWarmerFactory> policies,
+                                  final List<StdUserCacheWarmerFactory> users) {
+        this.policies = policies;
```
developer01189998819991197253, line 55: use setters created below

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdUserCacheWarmerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+    private Set<String> auths;
+    private Set<String> roles;
+
+    /**
+     * Constructor with 0 arguments for a standard implementation
+     * of the {@link UserCacheWarmerFactory} interface.
+     */
+    public StdUserCacheWarmerFactory() {
+        this.userId = "";
+        this.auths = Collections.emptySet();
```
developer01189998819991197253, line 45: use setters

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdUserConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Constructor with 1 argument for a standard implementation
+     * of the {@link UserConfiguration} interface
+     *
+     * @param users     a list of objects implementing the {@link uk.gov.gchq.palisade.service.UserCacheWarmerFactory} interface
+     */
+    public StdUserConfiguration(final List<StdUserCacheWarmerFactory> users) {
+        this.users = users;
```
developer01189998819991197253, line 51:
```suggestion
        this.setUsers(users);
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
```diff
@@ -48,9 +52,15 @@
...
     private final PolicyService service;
+    private PolicyConfiguration policyConfig;
+    private UserConfiguration userConfig;
 
-    public PolicyController(final @Qualifier("controller") PolicyService service) {
+    public PolicyController(final @Qualifier("controller") PolicyService service,
+                            final @Qualifier("policyConfiguration") PolicyConfiguration policyConfig,
+                            final @Qualifier("userConfiguration") UserConfiguration userConfig) {
         this.service = service;
+        this.policyConfig = policyConfig;
```
developer01189998819991197253, line 83: use setters

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
```diff
@@ -93,4 +103,12 @@ public void setTypePolicyAsync(@RequestBody final SetTypePolicyRequest request)
...
         service.setTypePolicy(request.getType(), request.getPolicy());
     }
+
+    @EventListener(ApplicationReadyEvent.class)
+    public void initPostConstruct() {
+        // Add example Policies to the policy-service cache
+        policyConfig.getPolicies().stream()
+                .map(cacheWarmer -> cacheWarmer.policyWarm(userConfig.getUsers()))
+                .forEach(entry -> service.setResourcePolicy(entry.getKey(), entry.getValue()));
+    }
```
developer01189998819991197253, line 144: add toString, hashCode and equals methods

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+    private String userId;
+    private Set<String> auths;
+    private Set<String> roles;
+
+    /**
+     * Constructor with 0 arguments for a standard implementation
+     * of the {@link UserCacheWarmerFactory} interface.
+     */
+    public StdUserCacheWarmerFactory() {
+        this.userId = "";
```
developer01189998819991197253, line 44: Use setters

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserCacheWarmerFactory.java
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * Constructor with 3 arguments for a standard implementation
+     * of the {@link UserCacheWarmerFactory} interface.
+     *
+     * @param userId    a {@link String} value of a user.
+     * @param roles     a {@link Set} of {@link String} role values for the user.
+     * @param auths     a {@link Set} of {@link String} auth values for the user.
+     */
+    public StdUserCacheWarmerFactory(final String userId, final Set<String> roles, final Set<String> auths) {
+        this.userId = userId;
```
developer01189998819991197253, line 58: use setters

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Constructor with 1 argument for a standard implementation
+     * of the {@link UserConfiguration} interface
+     *
+     * @param users     a list of objects implementing the {@link uk.gov.gchq.palisade.service.UserCacheWarmerFactory} interface
+     */
+    public StdUserConfiguration(final List<StdUserCacheWarmerFactory> users) {
+        this.users = users;
```
developer01189998819991197253, line 51: use setters

developer01189998819991197253 (changes requested):

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/service/PolicyServiceTest.java
```diff
@@ -110,7 +110,6 @@ public void infoOnGetPolicyRequest() {
         List<String> infoMessages = getMessages(event -> event.getLevel() == Level.INFO);
 
         MatcherAssert.assertThat(infoMessages, Matchers.hasItems(
-                Matchers.containsString(request.toString()),
```
dev930018, line 113: Why?
It's more worth checking the request has been logged than the response.
(Disregard if out-of-scope and failing, but raise a bug?)

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/ApplicationConfiguration.java
```diff
@@ -40,10 +42,39 @@
...
 
+    @Bean
+    @ConditionalOnProperty(prefix = "cache", name = "policyConfig", havingValue = "stdPolicyConfig")
+    public StdPolicyConfiguration policyConfiguration() {
+        LOGGER.info("Standard Policy Configuration Instantiated");
+        return new StdPolicyConfiguration();
+    }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "cache", name = "policyWarmer", havingValue = "stdPolicyCacheWarmer")
```
dev930018: Would this stdPolicyCacheWarmer ever change independently of the stdPolicyConfig?
Would it be better to have a `cache.policy: std` sort-of-thing if a stdPolicyWarmer <=> stdPolicyConfig

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/ApplicationConfiguration.java
```diff
@@ -40,10 +42,39 @@
...
+
+    @Bean
+    @ConditionalOnProperty(prefix = "cache", name = "userConfig", havingValue = "stdUserConfig")
+    public StdUserConfiguration userConfiguration() {
+        LOGGER.info("Standard User Configuration Instantiated");
+        return new StdUserConfiguration();
+    }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "cache", name = "userWarmer", havingValue = "stdUserCacheWarmer")
```
dev930018: ditto `cache.user: std`

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
dev930018: Hmm, I'd say don't use them, the getters and setters provide the public interface. If i removed the setter for resourceRules, i'd stilll want the constructor to work

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
```diff
@@ -0,0 +1,262 @@
...
+                policy.recordLevelRule(key, createRule(recordRules.get(key), "record"));
+            } catch (Exception ex) {
+                LOGGER.error("Error creating recordLevel Rule: {}", ex.getMessage());
+            }
+        }
+        return new MapEntry<>(createResource(), policy);
+    }
+
+    private <T> Rule<T> createRule(final String rule, final String ruleType) {
+        if ("resource".equalsIgnoreCase(ruleType)) {
```
dev930018, line 146: At what point do you make a Map of Exceptions to error messages? Probably now-ish. More extensible and reusable
`private static final Map<Exception, String> ERROR_MESSAGES = ...`
then
`LOGGER.error(ERROR_MESSAGES.get(ex), ex)`
(apparently just using the error with no "{}" in your error string is the 'proper' way to do this, e.g. `LOGGER.error("Error getting from class: ", ex)`)

dev930018: Might even be tempted to just have a general `catch (Exception ex)` "There was an error"

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
```diff
@@ -0,0 +1,262 @@
...
+        URI normalised = FileUtil.convertToFileURI(resource);
+        String resource = normalised.toString();
+        if (resource.endsWith(".avro")) {
+            return new FileResource().id(resource).type(type).serialisedFormat("avro").parent(getParent(resource));
+        } else {
+            return new DirectoryResource().id(resource).parent(getParent(resource));
+        }
+    }
+
+    private ParentResource getParent(final String fileURL) {
```
dev930018, line 176: We should standardise this stuff in common

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+     * of the {@link PolicyConfiguration} interface
+     */
+    public StdPolicyConfiguration() {
+    }
+
+    /**
+     * Constructor with 2 arguments for a standard implementation
+     * of the {@link PolicyConfiguration} interface
+     *
+     * @param policies  a {@link List} of objects implementing the {@link uk.gov.gchq.palisade.service.PolicyCacheWarmerFactory} interface
```
dev930018: qualified import `{@link PolicyCacheWarmerFactory}`

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param users  a {@link List} of objects implementing the {@link uk.gov.gchq.palisade.service.UserCacheWarmerFactory} interface
```
dev930018: ditto

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/ApplicationConfiguration.java
dev930018: Furthermore, looking further down, we're using`population` as the tag for the content.
This stuff doesn't really have anything to do with `cache`, more like `population.config` or something

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdUserConfiguration.java
dev930018: I'd disagree

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
dev930018: disagree

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
dev930018: Do we need them? Are we ever going to print a Controller? or compare one Controller to another?
I'd leave them out to avoid someone trying to do something weird

ac74475: I don't think the hashcode and equals is required here, but the toString could be useful

services-manager/src/main/resources/application-eureka.yaml
```diff
@@ -17,3 +17,31 @@
 eureka:
   client:
     enabled: true
+
```
dev930018: Unfortunately, this will probably cause problems when trying to start the palisade service with `eureka,h2` (or `eureka,jvm`

services-manager/src/main/resources/application-example.yaml
dev930018: Addressed in PAL-606 or 607... whatever I'm doing

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
pd104923: Setters and Getters added.

services-manager/src/main/resources/application-services.yaml
```diff
@@ -19,64 +19,37 @@
 # Extract the project version (x.y.z-REVISION etc.)
 app:
   version: '@project.version@'
-# While the library happens to be the same version as services, it should be declared here as a static name
```
dev930018, line 22: Since the default use of the manager is for running services for the example, why add the example profile?

services-manager/src/main/resources/application-example.yaml
```diff
@@ -0,0 +1,83 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+app:
+  version: '@project.version@'
+# While the library happens to be the same version as services, it should be declared here as a static name
+library:
+  version: ${app.version}
+
+manager:
```
dev930018, line 21: See comment as to `application.yaml`
But also, why reiterate the jar path when it's already in the default yaml?

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
pd104923: toString added.

policy-service/src/main/resources/application.yaml
```diff
@@ -48,3 +48,9 @@ management:
         - loggers
     loggers:
       enabled: true
+
+cache:
+  policyWarmer: stdPolicyCacheWarmer
```
dev930018: See comment on ApplicationConfiguration...
can this become:
```
populationConfig:
    policy: std
    user: std
```
?

services-manager/src/main/resources/application-discovery.yaml
```diff
@@ -27,5 +27,7 @@ manager:
   services:
     discovery-service:
       jar: discovery-service/target/discovery-service-${app.version}-exec.jar
+      profiles:
+        - discovery
```
dev930018: Why start the discovery service in discovery profile?

services-manager/src/main/resources/application-eureka.yaml
dev930018: Lists may not get amended with overlapping profiles, just straight overwritten

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+     * of the {@link UserConfiguration} interface
+     */
+    public StdUserConfiguration() {
+    }
+
+    /**
+     * Constructor with 1 argument for a standard implementation
+     * of the {@link UserConfiguration} interface
+     *
+     * @param users     a list of objects implementing the {@link uk.gov.gchq.palisade.service.UserCacheWarmerFactory} interface
```
dev930018: Qualified import `{@link UserCacheWarmerFactory}`

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
dev930018: Disagree

dev930018: Also, may have forgotten to hit unresolve on alot of these

user-service/src/main/java/uk/gov/gchq/palisade/service/user/web/UserController.java
```diff
@@ -18,34 +18,31 @@
...
 @RestController
 @RequestMapping(path = "/")
 public class UserController {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UserController.class);
     private UserService service;
+    private UserConfiguration userConfig;
 
-    public UserController(@Qualifier("userService") final UserService service) {
+    public UserController(@Qualifier("userService") final UserService service, @Qualifier("userConfiguration") final UserConfiguration configuration) {
```
dev930018, line 43: Why the qualifiers? are there `UserConfiguration`s other than the `@Bean("userConfiguration")`?

user-service/src/main/java/uk/gov/gchq/palisade/service/user/web/UserController.java
```diff
@@ -60,22 +57,11 @@ public Boolean addUserRequest(@RequestBody final AddUserRequest request) {
         return service.addUser(request.user).equals(request.user);
     }
 
-    @PostConstruct
+    @EventListener(ApplicationReadyEvent.class)
```
dev930018, line 60: Pros and cons vs PostConstruct? Readiness of spring beans and autowiring stuff i presume?

user-service/src/main/resources/application.yaml
```diff
@@ -60,3 +60,7 @@ feign:
 
 user-service:
   implementation: null-user-service
+
```
dev930018, line 63: ditto some other service yaml

dev930018 (changes requested):

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/web/PolicyController.java
dev930018: Agreed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/web/UserController.java
pd104923: Turns out that methods that have the @Chacheable annotation have no effect when used with an @PostConstruct annotated method.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/web/UserController.java
pd104923: On start up Spring creates a StdUserConfiguration bean as it is within the config package of the service. This helps the service know which to use in this situation, especially when starting in the example profile.

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/service/PolicyServiceTest.java
pd104923: This was removed as it was causing an issue, will raise a bug to adjust it in the future.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
pd104923: I can add this to the FileUtil class that I added in common to help standardise it.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
pd104923: This has already been addressed with a single Exception code block added (updated after being raised as a code smell by SonarQube)

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/ApplicationConfiguration.java
pd104923: This has now been updated to:
`@ConditionalOnProperty(prefix = "population", name = "policy", havingValue = "std")`
 for both Policy beans.
Change has also been made on the User beans.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/ApplicationConfiguration.java
pd104923: Amended to `population.user: std`, matching the policy-service update.

user-service/src/main/resources/application.yaml
pd104923: Updated

pd104923:
All constructor methods that were using a setter method have been reverted back to the previous process.

policy-service/src/main/resources/application.yaml
pd104923: Updated

user-service/src/main/java/uk/gov/gchq/palisade/service/user/config/StdUserConfiguration.java
pd104923: Updated

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyConfiguration.java
pd104923: Updated

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyConfiguration.java
pd104923: Updated.

services-manager/src/main/resources/application-discovery.yaml
pd104923: Not sure why I added that. It has now been removed.

services-manager/src/main/resources/application-eureka.yaml
pd104923: Confirmed, lists are not amended they are overwritten. Reverted back.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/config/StdPolicyCacheWarmerFactory.java
pd104923: @ac74475 would you agree with moving this over to common or would you prefer to leave it for now?

ac74475: I would leave this be for now. there is a ticket to tidy this stuff up later (PAL-612)

user-service/src/main/java/uk/gov/gchq/palisade/service/user/web/UserController.java
dev930018: I sense this one may have driven you mad for a while...

services-manager/src/main/resources/application-services.yaml
dev930018: Reasonable to expect either method

services-manager/src/main/resources/application-example.yaml
dev930018: Happy to have example profile, but cot down on duplication - we only really need the loader path stuff

services-manager/src/main/resources/application-example.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+  services:
+    audit-service:
+      jar: audit-service/target/audit-service-${app.version}-exec.jar
+      paths: *libraries
+      profiles:
+      - eureka
+      - example
+      log: audit-service-example.log
+      err: audit-service-example.err
```
dev930018, line 37: Final comment, promise!
Remove the duplicated jar, log, err fields. This should just enable the example stuff (so paths and profiles can stay)
And then ditto across other services here of course

dev930018 (dismissed):

services-manager/src/main/resources/application-example.yaml
ac74475: would you not want your example logs to go elsewhere? So this shows how you would do that

dev930018: Fair enough, i didn't notice the different name.
I'd still remove the jar field

dev930018: And then therefore the app.version bit at the top

developer01189998819991197253 (dismissed):

services-manager/src/main/resources/application-example.yaml
ac74475: So doing that change would mean that when starting the palisade example we need to have both services and example profiles. So I think for the sake of an extra line per service am happy to keep the jar's in here and decouple services (just loading the services) from the Example (loading everything needed for the example). Later on with the K8's smoke test stuff we will need to modify these to split out the eureka stuff as well anyway

ac74475 (dismissed):

services-manager/src/main/resources/application-example.yaml
pd104923: When starting the services for the example the services-manager uses the `example,debug` profiles. If this is removed then it would mean adding an extra profile to the startServices script.

ac74475 (approved):

PR #111, opened 2020-03-31
Pal 605 fix data service

nw1984:
Fix for resource service not being present and mount point

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -133,7 +133,8 @@ spec:
             container('maven') {
                 configFileProvider([configFile(fileId: "${env.CONFIG_FILE}", variable: 'MAVEN_SETTINGS')]) {
                     if (("${env.BRANCH_NAME}" == "develop") ||
-                            ("${env.BRANCH_NAME}" == "master")) {
+                            ("${env.BRANCH_NAME}" == "master")
+                                    ("${env.BRANCH_NAME}" == "PAL-605-fix-data-service")) {
                         sh 'palisade-login'
                         //now extract the public IP addresses that this will be open on
                         sh 'extract-addresses'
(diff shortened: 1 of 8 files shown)


Jenkinsfile
```diff
@@ -133,7 +133,8 @@ spec:
             container('maven') {
                 configFileProvider([configFile(fileId: "${env.CONFIG_FILE}", variable: 'MAVEN_SETTINGS')]) {
                     if (("${env.BRANCH_NAME}" == "develop") ||
-                            ("${env.BRANCH_NAME}" == "master")) {
+                            ("${env.BRANCH_NAME}" == "master") ||
+                                    ("${env.BRANCH_NAME}" == "PAL-605-fix-data-service")) {
```
developer01189998819991197253: after approved, please revert this back to original

developer01189998819991197253 (dismissed):

data-service/mvn_dependency_tree.txt
```diff
@@ -1,4 +1,5 @@
 uk.gov.gchq.palisade:data-service:jar:0.4.0-SNAPSHOT
++- com.google.code.gson:gson:jar:2.8.5:compile
```
ac74475, line 2: what was the reason for changing the gson version back a bug fix? (2.8.6 -> 2.8.5)

developer01189998819991197253: 2.8.6 doesnt seem to fix the issues, yet 2.8.5 does!

ac74475 (dismissed):

data-service/mvn_dependency_tree.txt
nw1984: Hi ac74475, the shaded hadoop client has a dependency on gson version <= 2.8.5 - there is an interface: com.google.gson.GsonBuilder com.google.gson.GsonBuilder.setLenient() that has been deleted from 2.8.6 onwards. The gson construction requires this hence the reason that we have to include the specific version.

ac74475 (approved):

developer01189998819991197253 (approved):

PR #119, opened 2020-04-23
updated Jenkinsfile to build Palisade services, then integration test…

developer01189998819991197253:
…s, and then upload to sonar with the joined code coverage

Code coverage went up to 55% from 47%
Doesnt show any extra unit tests but that might be a limitiation of our segregated repositories.
When only running integration tests it wont break if services do not exist in the correct folder heirarchy


dev-958 (dismissed):
Hard to validate just looking at the code but can't see any problems.

ac74475 (approved):

PR #124, opened 2020-04-29
PAL-646 improved jenkinsfiles

dev930018:


Jenkinsfile
```diff
@@ -72,39 +71,69 @@ spec:
...
-            //Repositories must get built in their own directory, they can be 'cd' back into later on
+
+        stage('Install, Unit Tests, Checkstyle') {
             dir ('Palisade-services') {
                 git url: 'https://github.com/gchq/Palisade-services.git'
                 sh "git fetch origin develop"
-                // CHANGE_BRANCH will be null unless you are building a PR, in which case it'll become your original branch name, i.e pal-xxx
-                // If CHANGE_BRANCH is null, git will then try to build BRANCH_NAME which is pal-xxx, and if the branch doesnt exist it will default back to develop
-                sh "git checkout ${env.CHANGE_BRANCH} || git checkout ${env.BRANCH_NAME} || git checkout develop"
+                sh "git checkout ${GIT_BRANCH_NAME} || git checkout develop"
```
nw1984: Helpful to echo here which branch name is being built.

Jenkinsfile
```diff
@@ -72,39 +71,69 @@ spec:
...
+
         stage('Integration Tests') {
             dir ('Palisade-integration-tests') {
-            git url: 'https://github.com/gchq/Palisade-integration-tests.git'
-            sh "git fetch origin develop"
-            // CHANGE_BRANCH will be null unless you are building a PR, in which case it'll become your original branch name, i.e pal-xxx
-            // If CHANGE_BRANCH is null, git will then try to build BRANCH_NAME which is pal-xxx, and if the branch doesnt exist it will default back to develop
-            sh "git checkout ${env.CHANGE_BRANCH} || git checkout ${env.BRANCH_NAME} || git checkout develop"
+                git url: 'https://github.com/gchq/Palisade-integration-tests.git'
+                sh "git fetch origin develop"
```
nw1984: Don't think we need to fetch origin develop - nothing should have changed since we set the url and did the git checkout either on the brach or develop

Jenkinsfile
```diff
@@ -72,39 +71,69 @@ spec:
...
                     }
                 }
             }
         }
-        stage('SonarQube analysis') {
+
+        stage('SonarQube Analysis') {
             dir ('Palisade-services') {
                 container('docker-cmds') {
                     withCredentials([string(credentialsId: '3dc8e0fb-23de-471d-8009-ed1d5890333a', variable: 'SONARQUBE_WEBHOOK'),
```
nw1984, line 140: Credentials are stored in a environmental parameter in the new infrastructure. We should do the same on the old and use these parameters rather than this exact id.

Jenkinsfile
```diff
@@ -142,7 +180,7 @@ spec:
                             //now extract the public IP addresses that this will be open on
                             sh 'extract-addresses'
                             sh 'mvn -s $MAVEN_SETTINGS deploy -Dmaven.test.skip=true'
-                            sh 'helm upgrade --install palisade . --set traefik.install=true,dashboard.install=true   --set global.repository=${ECR_REGISTRY}  --set global.hostname=${EGRESS_ELB} --set global.localMount.enabled=false,global.localMount.volumeHandle=${VOLUME_HANDLE} --namespace dev'
+                            sh 'helm upgrade --install palisade . --set traefik.install=true,dashboard.install=true --set global.repository=${ECR_REGISTRY} --set global.hostname=${EGRESS_ELB} --set global.localMount.enabled=false --set global.localMount.volumeHandle=${VOLUME_HANDLE} --namespace dev'
```
nw1984: I think get rid of all the set's and just use a comma separated list - makes it a bit more readable.

Jenkinsfile
```diff
@@ -72,39 +71,69 @@ spec:
          path: /var/run
 ''') {
     node(POD_LABEL) {
+        def GIT_BRANCH_NAME
```
nw1984, line 74: default to env.BRANCH_NAME and then you can get rid of the else section on line 79--81.

nw1984 (changes requested):

Jenkinsfile
dev930018: I feel like the behaviour is more obvious if its all in the if/else. It shows its clearly not used for anything before the if/else that way

Jenkinsfile
dev930018: Out-of-scope?

nw1984 (dismissed):

Jenkinsfile
```diff
@@ -72,39 +71,66 @@ spec:
...
         stage('Integration Tests') {
             dir ('Palisade-integration-tests') {
-            git url: 'https://github.com/gchq/Palisade-integration-tests.git'
-            sh "git fetch origin develop"
-            // CHANGE_BRANCH will be null unless you are building a PR, in which case it'll become your original branch name, i.e pal-xxx
-            // If CHANGE_BRANCH is null, git will then try to build BRANCH_NAME which is pal-xxx, and if the branch doesnt exist it will default back to develop
-            sh "git checkout ${env.CHANGE_BRANCH} || git checkout ${env.BRANCH_NAME} || git checkout develop"
+                git url: 'https://github.com/gchq/Palisade-integration-tests.git'
+                sh "git checkout ${GIT_BRANCH_NAME} || git checkout develop"
+                echo sh(script: 'git rev-parse --abbrev-ref HEAD', returnStdout: true)
```
developer01189998819991197253: why is this needed?

developer01189998819991197253: specifically the git rev parse

dev930018: `git reverse-parse --abbreviated-reference HEAD -> PAL-xxx-branch-name`
@nw1984 suggested printing out exactly which branch is being used, just for clarity (see above comment)

developer01189998819991197253: when the sh git checkout line runs it will output "checked out x branch" 
so there will be logs available. we either need to do it for all git checkouts or none

dev930018: Yeah, i was just looking at jenkins, the branch checked out does get printed anyway

developer01189998819991197253: for example, pulling from pal-515 jenkinsfile logs you get:

```
+ git checkout pal-515-JVM-Smoke-Tests
Branch 'pal-515-JVM-Smoke-Tests' set up to track remote branch 'pal-515-JVM-Smoke-Tests' from 'origin'.
Switched to a new branch 'pal-515-JVM-Smoke-Tests'
```

developer01189998819991197253 (approved):

ac74475 (approved):

PR #128, opened 2020-05-18
Pal 642 data service prepopulation

pd104923:


data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserConstructorNotFoundException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package uk.gov.gchq.palisade.service.data.exception;
+
+/**
+ * A {@link RuntimeException} that is thrown when incorrect values are used for the service Prepopulation.
+ */
+public class SerialiserConstructorNotFoundException extends RuntimeException {
+
+    /**
```
m78233, line 24: Would have expected the other two constructor methods that we look like we use by convention:
SerialiserConstructorNotFoundException(final String message)
SerialiserConstructorNotFoundException(final Throwable cause)

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserInitialisationException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class SerialiserInitialisationException extends RuntimeException {
+
+    /**
+     * Constructs a new {@link SerialiserInitialisationException} with the specified detail message and cause.
+     *
+     * @param message   a {@link String} value detailing the error
+     * @param cause     a {@link Throwable} that caused the iss
+     */
+    public SerialiserInitialisationException(final String message, final Throwable cause) {
```
m78233, line 48: Similar question here as with SerialiserConstructorNotFoundException,  consider the other constructors.

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserNotFoundException.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public class SerialiserNotFoundException extends RuntimeException {
+
+    /**
+     * Constructs a new {@link SerialiserNotFoundException} with the specified detail message and cause.
+     *
+     * @param message   a {@link String} value detailing the error
+     * @param cause     a {@link Throwable} that caused the iss
+     */
+    public SerialiserNotFoundException(final String message, final Throwable cause) {
```
m78233, line 48: Similar question here as with SerialiserConstructorNotFoundException, consider the other constructors.

developer01189998819991197253 (dismissed):

data-service/src/main/java/uk/gov/gchq/palisade/service/data/web/DataController.java
```diff
@@ -53,10 +67,24 @@ public DataController(final DataService service) {
     @PostMapping(value = "/addSerialiser", consumes = "application/json", produces = "application/json")
```
ac74475: given we are now using pre population, I don't think we need to have a public rest end point for adding the serialisers, which means we can remove the addSerialiserRequest class. I would leave the data service interface addSerialiser method though

ac74475 (changes requested):
One small change to remove some unnecessary parts

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserConstructorNotFoundException.java
```diff
@@ -0,0 +1,51 @@
...
+     * Constructs a new {@link SerialiserConstructorNotFoundException} with the specified detail message and cause.
```
developer01189998819991197253: change it so javadoc is specific for each constructor. 
One for message
One for cause
one for both, same in SerlaiserInitialisationException and SerialiserNotFoundException

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserInitialisationException.java
```diff
@@ -0,0 +1,51 @@
...
+
+package uk.gov.gchq.palisade.service.data.exception;
+
+/**
+ * A {@link RuntimeException} that is thrown when incorrect values are used for the service Prepopulation.
+ */
+public class SerialiserInitialisationException extends RuntimeException {
+
+    /**
+     * Constructs a new {@link SerialiserInitialisationException} with the specified detail message and cause.
```
developer01189998819991197253: change it so javadoc is specific for each constructor. 
One for message
One for cause
one for both, same in SerlaiserInitialisationException and SerialiserNotFoundException

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserNotFoundException.java
```diff
@@ -0,0 +1,51 @@
...
+     * Constructs a new {@link SerialiserNotFoundException} with the specified detail message and cause.
+     *
+     * @param cause     a {@link Throwable} that caused the error
+     */
+    public SerialiserNotFoundException(final Throwable cause) {
+        super(cause);
+    }
+
+    /**
+     * Constructs a new {@link SerialiserNotFoundException} with the specified detail message and cause.
```
developer01189998819991197253: change it so javadoc is specific for each constructor. 
One for message
One for cause
one for both, same in SerlaiserInitialisationException and SerialiserNotFoundException

developer01189998819991197253 (changes requested):

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserNotFoundException.java
pd104923: Fixed

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserInitialisationException.java
pd104923: Fixed

data-service/src/main/java/uk/gov/gchq/palisade/service/data/exception/SerialiserConstructorNotFoundException.java
pd104923: Fixed

data-service/src/main/java/uk/gov/gchq/palisade/service/data/web/DataController.java
pd104923: The AddSerialiserRequest has been deleted and the rest end point has been removed.

pd104923: These have been removed, required some changes to the integration tests.

m78233 (dismissed):

ac74475 (dismissed):

developer01189998819991197253 (dismissed):

dev930018 (approved):

dev930018 (changes requested):
You will need changes in all resource services to use fully qualified classpath for the type
For the simple resource service, i would recommend using java.lang.String as the type
To avoid having to change anything for hadoop, you should just be able to rename the employee_fileXX.avro to uk.gov.....Employee_fileXX.avro
This should be done in all appropriate places

ac74475:
I don't like the idea of changing the name of the files, I would rather have a lookup map of known types within the HadoopResourceDetails class which can map employee to the fully qualified domain name for the Employee class. I agree with changing the currently hardcoded value of 'txt' too 'java.lang.String' in the SimpleResourceService.

dev930018 (approved):

m78233 (approved):

ac74475 (approved):

developer01189998819991197253 (approved):

PR #132, opened 2020-05-29
Pal 324 new infrastructure changes

nw1984:


Jenkinsfile
```diff
@@ -14,13 +14,53 @@
  * limitations under the License.
  */
 
+
+//node-affinity
+//nodes 1..3 are reserved for Jenkins slave pods.
+//node 0 is used for the Jenkins master
+//node 4 is used for SQ and Nexus
```
ac74475: I don't think we need to know that node 4 is used for SQ and Nexus. But knowing about Jenkins slaves and master are useful
```suggestion
```

Jenkinsfile
```diff
@@ -14,13 +14,53 @@
...
+    - $(JENKINS_NAME)
+    resources:
+      requests:
+        ephemeral-storage: "4Gi"
+      limits:
+        ephemeral-storage: "8Gi"
+
   - name: docker-cmds
-    image: jnlp-did:jdk11
+    image: 779921734503.dkr.ecr.eu-west-1.amazonaws.com/jnlp-did:INFRA
```
ac74475: Does this need to specify the ECR? If so can we parameterise this?

Jenkinsfile
```diff
@@ -29,62 +69,80 @@ spec:
...
-        value: ""
+    resources:
+      requests:
+        ephemeral-storage: "1Gi"
+      limits:
+        ephemeral-storage: "2Gi"
+
   - name: maven
-    image: jnlp-slave-palisade:jdk11
+    image: 779921734503.dkr.ecr.eu-west-1.amazonaws.com/jnlp-dood-new-infra:INFRA
```
ac74475: Does this need to specify the ECR? If so can we parameterise this?

Jenkinsfile
```diff
@@ -190,3 +237,70 @@ spec:
...
+//    volumeMounts:
+//      - name: docker-graph-storage
+//        mountPath: /var/lib/docker
+//
+//  volumes:
+//    - name: docker-graph-storage
+//      emptyDir: {}
+//    - name: docker-sock
+//      hostPath:
+//         path: /var/run
```
ac74475: are these required?

ac74475 (changes requested):

Jenkinsfile
nw1984: Since we are using a real Kubernetes installation rather than microk8s, it is necessary to specify the full tag for the ECR.
I have tried parameterising the variable using ${}, ${env.}, $(), $(env.) - none of which result in the substitution working at all.
The intention, is to incorporate the pod template definition into the Jenkins template longer term so that the Jenkinsfiles are sparse, so the visibility of the ECR registry should not remain an issue.
I can spend more time looking at this but it is not as obvious as it would first appear.

Jenkinsfile
nw1984: deleted.

Jenkinsfile
```diff
@@ -116,29 +173,30 @@ spec:
...
+//            dir ('Palisade-integration-tests') {
+//                git url: 'https://github.com/gchq/Palisade-integration-tests.git'
+//                sh "git checkout ${GIT_BRANCH_NAME} || git checkout develop"
+//                container('docker-cmds') {
+//                    configFileProvider([configFile(fileId: "${env.CONFIG_FILE}", variable: 'MAVEN_SETTINGS')]) {
+//                        sh 'mvn -s $MAVEN_SETTINGS install'
+//                    }
+//                }
+//            }
+//        }
```
ac74475: Will need to uncomment this section before this can be merged, but waiting on PAL-652

dev-958 (dismissed):

dev-958 (dismissed):

ac74475 (dismissed):

ac74475 (dismissed):

dev-958 (dismissed):

ac74475 (approved):

PR #135, opened 2020-06-05
Pal 641 create notices file

pd104923:


ac74475 (commented):
Looks really good, Makes it much easier to see when we have changed versions of dependancies or even have multiple versions. E.g caffeine and eureka.

ac74475 (changes requested):
Is it possible to remove the generation of the html licence files?

ac74475:
doesn't appear to be easily possible within the plugin as it downloads all the licences as per the link that the dependency stated its licence link to be. Therefore some projects have defined that as a .txt and some as .html files.

ac74475 (dismissed):

dev930018 (dismissed):

dev930018:
@ac74475 There has been a preference to link the notices.md to the txt license if available

data-service/pom.xml
```diff
@@ -26,11 +26,13 @@
...
     </parent>
 
+    <!--  *********** Aritfact ID ***********    -->
     <artifactId>data-service</artifactId>
+    <url>https://github.com/gchq/Palisade-services/tree/develop/data-service</url>
+    <!--  *********** Aritfact ID ***********    -->
 
     <properties>
         <eureka.version>1.9.16</eureka.version>
-        <databind.version>2.9.8</databind.version>
```
dev930018: Why has this gone?

policy-service/pom.xml
```diff
@@ -38,7 +41,7 @@
         <dependency>
             <groupId>com.github.ben-manes.caffeine</groupId>
             <artifactId>caffeine</artifactId>
-            <version>${caffeine.version}</version>
+            <version>2.8.1</version>
```
dev930018: Should be reverted back to caffeine.version

user-service/pom.xml
```diff
@@ -25,11 +25,15 @@
...
 
+    <!--  *********** Aritfact ID ***********    -->
     <artifactId>user-service</artifactId>
+    <url>https://github.com/gchq/Palisade-services/tree/develop/user-service</url>
+    <!--  *********** Aritfact ID ***********    -->
 
     <properties>
         <spring-cloud-starter-ribbon.version>1.4.7.RELEASE</spring-cloud-starter-ribbon.version>
         <swagger.version>1.5.21</swagger.version>
+        <eureka.version>1.9.16</eureka.version>
```
dev930018: No point adding this as it's getting removed in PAL-653

user-service/pom.xml
```diff
@@ -75,6 +79,7 @@
         <dependency>
             <groupId>com.netflix.eureka</groupId>
             <artifactId>eureka-client</artifactId>
+            <version>${eureka.version}</version>
```
dev930018: ditto, revert

dev930018 (changes requested):

developer01189998819991197253 (commented):
Approved after previous comments are addressed

data-service/pom.xml
pd104923: There was a duplication of the Jackson Databind license information. This value was the older version and was not being used within the pom so I removed it.

user-service/pom.xml
pd104923: Will remove this.

policy-service/pom.xml
pd104923: This was using version 2.8.0 from the `spring-boot-starter-parent 2.2.0.RELEASE`. Added 2.8.1 as a property in the pom to override this.

pd104923: This has also been done in the user-service pom as well.

user-service/pom.xml
pd104923: Reverted

m78233 (approved):

dev930018 (dismissed):

nw1984 (dismissed):

m78233 (dismissed):

ac74475 (dismissed):

developer01189998819991197253 (dismissed):

ac74475 (approved):

PR #138, opened 2020-06-17
PAL-682 services develop failing helm install

dev930018:
You can validate it by doing
`helm upgrade --install palisade . --set hosting=aws --set traefik.install=true,dashboard.install=true --set global.repository=ecr,global.hostname=elb --set global.persistence.classpathJars.aws.volumeHandle=classpaths,global.persistence.dataStores.palisade-data-store.aws.volumeHandle=data,global.persistence.kafka.aws.volumeHandle=kafka,global.persistence.redis.aws.volumeHandle=redis --namespace dev --dry-run`
which is what is in the jenkinsfile, substituting out the secrets for dummy placeholders
Piping it through `less` you can inspect the output for the data-service PVs, PVCs and Deployment and it all looks correct.
The original cause was all from the jenkinsfile doing a `--set global.persistence.dataStores[0]...` which is no longer there, dataStores has been changed from a list to a map
It wasn't picked up in testing because in the local profile, the `--set` covered all (one) of the tags (`local.hostPath`), whereas aws covered half (one of two) of them (`aws.volumeHandle` was set, which erased `aws.volumePath`)


pd104923 (dismissed):

templates/data-store-pv.yaml
```diff
@@ -12,37 +12,37 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Keep track of the root . inside control structures
-{{- $root := . }}
+# Keep track of the 'hosting' tag inside control structures, as '.' is overridden and '.Values' is inaccessible
+{{- $hosting := .Values.hosting }}
```
dev-958: this variable doesn't seem to serve any purpose, if it is not to be mutated use the .Values notation.

values.yaml
```diff
@@ -87,14 +87,15 @@ global:
...
     dataStores:
-    - name: palisade-local-data-store
-      mountPath: "/data/local-data-store"
-      storage: 1Gi
-      local:
-        hostPath: "/"
-      aws:
-        volumeHandle: null
-        volumePath: "/"
+      # nb. Helm 3 has difficulties with --set and lists, use a map here instead
```
dev-958: please remove this comment

dev-958 (changes requested):

templates/data-store-pv.yaml
dev930018: The range control structure overrides the value of `.` to whatever range item you are operating on, making `.Values` point to (in this case) `global.persistence.dataStore.palisade-data-store.Values`

dev930018: Changed to $, t.i.l.

dev-958 (approved):

ac74475 (approved):

PR #140, opened 2020-06-23
Pal 683 fix helm deployment issue

nw1984:


charts/data-service/templates/deployment.yaml
```diff
@@ -67,21 +67,18 @@ spec:
         volumeMounts:
         - name: {{ .Values.global.persistence.classpathJars.name }}
           mountPath: {{ .Values.global.persistence.classpathJars.mountPath }}/{{ .Chart.Name }}/classpath/{{ .Chart.Version }}
-        {{- range $name, $value := .Values.global.persistence.dataStores }}
-        - name: {{ $name }}
-          mountPath: {{ $value.mountPath }}
-        {{- end }}
+
```
dev-958: These still need to be processed as a range. There will need to be a mount for every source data-store that a data service can access.

charts/data-service/templates/deployment.yaml
```diff
@@ -67,21 +67,18 @@ spec:
...
+        - name: {{ .Values.global.persistence.dataStores.name }}
+          mountPath: {{ .Values.global.persistence.dataStores.mountPath }}
       volumes:
       - name: {{ .Values.global.persistence.classpathJars.name }}
         persistentVolumeClaim:
           claimName: {{ .Values.global.persistence.classpathJars.name }}
           readOnly: true
-      {{- range $name, $_ := .Values.global.persistence.dataStores }}
-      - name: {{ $name }}
+      - name: {{ .Values.global.persistence.dataStores.name }}
```
dev-958: ditto

charts/data-service/templates/deployment.yaml
```diff
@@ -67,21 +67,18 @@ spec:
...
         persistentVolumeClaim:
-          claimName: {{ $name }}
+          claimName: {{ .Values.global.persistence.dataStores.name }}
```
dev-958: ditto

templates/data-store-pv.yaml
```diff
@@ -14,32 +14,37 @@
 
 # Create a number of persistent volumes for each configured data-store
 # New PV for each dataStores entry
-{{- range $name, $value := .Values.global.persistence.dataStores }}
```
dev-958, line 17: Again, these need to be a range.

templates/data-store-pvc.yaml
```diff
@@ -14,21 +14,29 @@
 
 # Create a number of persistent volumes for each configured data-store
 # New PVC for each dataStores entry
-{{- range $name, $value := .Values.global.persistence.dataStores }}
```
dev-958, line 17: again range...

templates/storageclass.yaml
```diff
@@ -0,0 +1,7 @@
+{{- if and ($.Values.global.persistence.storageClassDeploy) (eq $.Values.hosting "aws") }}
```
dev-958, line 1: Not sure this is needed like this (storageClassDeploy) as the SC would only be redeployed by helm if it was changed.

values.yaml
```diff
@@ -65,6 +65,8 @@ global:
 
   # persistent-volume and associated claims configurations
   persistence:
+    storageClassDeploy: false
```
dev-958, line 68: I don't think you need this variable

values.yaml
```diff
@@ -87,14 +89,14 @@ global:
     # dataStores: List of stores used by data-service and associated data-readers to read resources and return to a client
     # nb. the mountPath here will be referenced by all client resource requests
     dataStores:
-      palisade-data-store:
```
dev-958: The concept here is that the data mounts are a map so that a given service can mount more than one storage if required.

dev-958 (changes requested):

templates/storageclass.yaml
nw1984: The storage class changes are normally done at the point of EKS pipeline creation at the point of the efs csi driver instantiation. I wanted to include the definition here so that if a deployment was not done correctly, the user was aware of the changes that might be needed.

values.yaml
nw1984: The storage class changes are normally done at the point of EKS pipeline creation at the point of the efs csi driver instantiation. I wanted to include the definition here so that if a deployment was not done correctly, the user was aware of the changes that might be needed.

ac74475 (commented):
Can check and update where required the licence headers on the charts that you have made changes to that we previously just copies from other repos. If we have made changes then we need to add the crown copyright and indicate that it was 'Originally copied from ...' rather than 'Copied from ...'

pd104923 (approved):

ac74475 (changes requested):
Am not keen on the master-slave terminology, I know we are inheriting it from the Redis charts but I would like to see all references to 'slave' be replaced with 'worker'. This will almost certainly mean we need to update the licence header on a bunch of the Redis chart files.

dev-958 (changes requested):
Picking up the wrong storage class on local install:
image

charts/redis/templates/redis-master-statefulset.yaml
```diff
@@ -28,6 +30,30 @@ spec:
...
+      - ReadWriteMany
+      resources:
+        requests:
+          storage: {{ .Values.master.persistence.size | quote }}
+      storageClassName: {{ .Values.global.persistence.storageClassName }}
+      selector:
+        matchLabels:
+          selectorFilter: {{ .Values.global.persistence.redisMaster.name }}
+      {{- if .Values.master.persistence.matchExpressions }}
+      matchExpressions: {{- toYaml .Values.master.persistence.matchExpressions | nindent 12 }}
```
dev-958, line 53: I think this may need an extra tab space...

dev-958 (changes requested):

charts/redis/templates/redis-master-statefulset.yaml
nw1984: Discussion with @dev-958 - that developer has agreed to make the changes necessary.

ac74475 (approved):

dev-958 (approved):

PR #142, opened 2020-06-24
updated hadoop version and built licenses

developer01189998819991197253:


ac74475 (changes requested):
Add the licence files back in

ac74475 (approved):

PR #145, opened 2020-06-30
Pal 423 new message types

m78233:


audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/request/AuditMessage.java
```diff
@@ -0,0 +1,230 @@
...
+                .add("numberOfRecordsReturned=" + numberOfRecordsReturned)
+                .add("numberOfRecordsProcessed=" + numberOfRecordsProcessed)
+                .add("errorMessage='" + errorMessage + "'")
+                .add(super.toString())
+                .toString();
+    }
+
+    /**
+     * Builder class for the creation of instances of the ResultsRequest.  This is a variant of the Fluent Builder
+     * which will use Java Objects or JsonNodes equivalents for the components in the build.
```
pd104923:
```suggestion
     * Builder class for the creation of ResultsRequest instances.  This is a variant of the Fluent Builder
     * which will use Java Objects or JsonNode equivalents for the components in the build.
```

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/request/OriginalRequest.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Objects;
+import java.util.StringJoiner;
+
+
+/**
+ * Represents the original data that has been sent from the client to Palisade Service for a request to access data.
+ * This data will be forwarded to a set of services with each contributing to the processing of this request.
+ * This version represents the original request.
+ * Next in the sequence is User Service where this data will be the request for a User.
+ * Note there are two class that represent effectively the same data but represent a different stage of the process.
```
pd104923:
```suggestion
 * Next in the sequence is User Service where this data will be used to request a User.
 * Note there are two classes that, effectively, represent the same data but represent a different stage of the process.
```

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/request/OriginalRequest.java
```diff
@@ -0,0 +1,142 @@
...
+
+    public String getUserId() {
+        return userId;
+    }
+
+    public String getResourceId() {
+        return resourceId;
+    }
+
+    public Context getContext() throws JsonProcessingException {
```
pd104923: These should probably have the "@ Generated" tags on them.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/response/AuditMessage.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    public final String timeStamp; //when the service processed the request
+    public final String serverIp;   //server for the service that processed the request
+    public final String serverHostname;  //server host name for the service
+    public final Context context;  //Context of the client's request
+    public final String userId;    //User Id for the client.  Can be null if there is a User
+    public final String resourceId;  //Resource Id for the client.  Can be null if there is a Resource
+    public final String errorMessage;  //Error message that occurred during the processing of the request.  Will be null if there was no issue.
```
pd104923, line 49: Missing some fields when compared to the audit-service AuditMessage class

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/response/ErrorResponse.java
```diff
@@ -0,0 +1,67 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.util.Assert;
+
+/**
+ * Error message in human readable form.  This can be generated in any of the services.  Once an error occurs
+ * in a service,  processing of the requests stops.  This messaging is constructed and forwarded to the Results
+ * Service skipping any services that have not been preformed.  Results services will forward this message back
+ * to client who should be given enough information to correct the problem before tying again.
+ * The technical information will contain information that may help in understanding the issue and so may contain
+ * information about the service which should not be made public such as the stack trace of the error.
```
pd104923:
```suggestion
 * Error message in human readable form.  This can be generated in any of the services.  Once an error occurs
 * in a service,  processing of the request stops.  This message is constructed and forwarded to the results-service
 * skipping any other services that have not been preformed.  The results-service will forward this message back
 * to client who should be given enough information to correct the problem before tying again.
 * The technical message will contain information that may help in understanding the issue and so may contain
 * information about the service which should not be made public such as the stack trace of the error.
```

policy-service-example.err
```diff
@@ -0,0 +1,5 @@
+WARNING: An illegal reflective access operation has occurred
```
pd104923: This file needs to be removed.

policy-service-example.log
```diff
@@ -0,0 +1,375 @@
+16:30:49.002 [main] DEBUG uk.gov.gchq.palisade.service.policy.PolicyApplication - PolicyApplication started with: class uk.gov.gchq.palisade.service.policy.PolicyApplication main []
```
pd104923: This file needs to be removed.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/request/PolicyRequest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public Context getContext() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.context, Context.class);
+    }
+
+    public User getUser() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.user, User.class);
+    }
+
+    public LeafResource getResource() throws JsonProcessingException {
```
pd104923: "@ Generated" tags needed for these methods?

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/AuditMessage.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Audit information for a request provided by each service.
+ * Each individual service sends a record to the Audit Service for every request that it receives.
+ * The components of the message will differ depending on which service has sent the data and if the processing was
+ * successful or not.
+ *
+ */
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public class AuditMessage {
```
pd104923, line 38: The AuditMessage for each service needs to match the message that the audit-service is expecting. Some of the fields can be null but surely they still need to be there to allow it to be deserialised properly.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/ErrorResponse.java
```diff
@@ -0,0 +1,67 @@
...
+        Assert.notNull(technicalMessage, "TechnicalMessage cannot be null");
+        Assert.notNull(errorMessage, "ErrorMessage cannot be null");
+
+        this.technicalMessage = technicalMessage;
+        this.errorMessage = errorMessage;
+    }
+
+
+
+    public static class Builder {
```
pd104923, line 56: The JavaDoc for this static class needs to be added.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/PolicyResponse.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public Context getContext() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.context, Context.class);
+    }
+
+    public User getUser() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.user, User.class);
+    }
+
+    public LeafResource getResource() throws JsonProcessingException {
```
pd104923: "@ Generated" tags?

query-scope-service/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>service</artifactId>
+        <groupId>uk.gov.gchq.palisade</groupId>
+        <version>0.4.0-${revision}</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>query-scope-service</artifactId>
```
pd104923, line 29: Can we add in the url tag like the other pom files. This will allow it to be added to any NOTICES files in the future. I will be a link along the lines of 'https://github.com/gchq/Palisade-services/tree/develop/query-scope-service'

query-scope-service/src/main/java/uk/gov/gchq/palisade/service/queryscope/request/QueryScopeRequest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    public User getUser() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.user, User.class);
+    }
+
+    public LeafResource getResource() throws JsonProcessingException {
+        return MAPPER.treeToValue(this.resources, LeafResource.class);
+    }
+
+    public Rules getRules() throws JsonProcessingException {
```
pd104923: "@ Generated" tags?

audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/request/AuditMessage.java
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright 2019 Crown Copyright
```
pd104923:
```suggestion
 * Copyright 2020 Crown Copyright
```
This will need to reviewed on the other AuditMessage files as well

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/request/AuditMessageTest.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright 2019 Crown Copyright
```
pd104923:
```suggestion
 * Copyright 2020 Crown Copyright
```
Also needs to be checked for the other AuditMessageTest files

results-service/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>service</artifactId>
+        <groupId>uk.gov.gchq.palisade</groupId>
+        <version>0.4.0-${revision}</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>results-service</artifactId>
```
pd104923, line 29: Same comment as query-scope-service (add the url tag)

user-service/src/main/java/uk/gov/gchq/palisade/service/user/response/ErrorResponse.java
```diff
@@ -0,0 +1,60 @@
...
+ * in a service,  processing of the requests stops.  This messaging is constructed and forwarded to the Results
+ * Service skipping any services that have not been preformed.  Results services will forward this message back
+ * to client who should be given enough information to correct the problem before tying again.
+ */
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+public class ErrorResponse {
+
+    //This version of the message may not necessarily be sanitised.  It will need to pass information to
+    // the Audit Service that may technical details as to the reason for the failure.  This can then be used to
+    // resolve the issue if it was technical in natured.
```
pd104923, line 33:
```suggestion
    // the Audit Service that may include technical details as to the reason for the failure.  This can then be used to
    // resolve the issue if it was technical in nature.
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/response/UserResponse.java
```diff
@@ -0,0 +1,126 @@
...
+        default IUser withContext(Map<String, String> context) {
+            return withContextNode(MAPPER.valueToTree(context));
+        }
+
+        IUser withContextNode(JsonNode context);
+
+    }
+
+
+    interface IUser {
```
pd104923: It looks like this has been added outside of the static Builder class. Should be inside the static class

user-service/src/main/java/uk/gov/gchq/palisade/service/user/response/UserResponse.java
```diff
@@ -0,0 +1,126 @@
...
+        interface IResource {
+            /**
+             * @param resourceId {@link String} is the resource id provided in the register request
+             * @return the {@link UserResponse}
+             */
+            IContext withResource(String resourceId);
+        }
+    }
+
+    interface IContext {
```
pd104923: It looks like this has been added outside of the static Builder class. Should be inside the static class

pd104923 (changes requested):
A few things to address:
Copyright year needs to be updated in all AuditMessage and AuditMessageTest classes.
Shouldn't the content of the AuditMessage classes be the same as what's expected in the AuditService (even if the field is going to have a null value)?
"@ Generated" tags should be added to all the getMethods within the Request and Response classes.
All *-example.err and *-example.log files need to be removed from this branch.
As discussed, to keep clarity on the review, any suggested changes to the comments in a file should also be checked/re-worked in the other Request and Response files in the branch.

There are also a large number of SQ smells that should probably be addressed as well.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/request/OriginalRequest.java
m78233: There was a different idea about getters that came out of the discussions.  If the attribute is meant to be used in the service, it is given public scope and so we go for the direct approach (userRequest.user ).  The attributes that are not expected to be used in the service are held as Strings (if they are really simple) or a JsonNode for an Object.  They are given a private scope.  In this context the getters are convenience methods for something that in theory we should not need.  Note for example, the getContext method changes the object from a JSonNode to a Context object which does not conform to the getter method template and so should not have the @Generated annotation.   If I use them for some getters and not others it could be confusing.

If anything there is an argument for not having getters at all since the private data is not expected to be used outside of the class.  I thought that we could debate later when people start working with them and see what is right and wrong with then.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/response/AuditMessage.java
m78233: The idea here is that each of the AuditMessage would only hold the data that is relevant to the service that sent the data (plus optionally the error message if there is one).  This means that the AuditMessage for the palisade service will only have to know about things like userID (a String) and resourceID (a String).  It does not need to know anything about User Resource, Rules etc and any baggage associated with other services.  The AuditMessage in the audit-service is at the other extreme of this and would have to know about all of the baggage for all of the different services as it would be getting data from each of the services.

policy-service-example.err
m78233: yes

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/AuditMessage.java
m78233: I think that the missing attributes are handled properly.  See the comments for the palisade-service version of AuditMessage and the Unit test that I have added to the audit-service.  The key point is each service is to only handle what it knows about and the audit-service will have to handle all of these variations.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/ErrorResponse.java
m78233: yes

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/response/PolicyResponse.java
m78233: Not sure.  See note for OriginalRequest.

query-scope-service/pom.xml
m78233: Sounds good.  Not familiar with the idea, but I will look it up and make the necessary changes.

query-scope-service/src/main/java/uk/gov/gchq/palisade/service/queryscope/request/QueryScopeRequest.java
m78233: Not sure. See note for OriginalRequest.

results-service/pom.xml
m78233: I'll look into it

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/response/AuditMessage.java
m78233: I had added a small unit test to the AuditMessageTest class (for audit-service).  It is a test where the starting point is a Json string representing a AuditMessage from the palisade-service and the object that is being used in the deserialising is the AuditMessage for the audit-service.  This is meant to represent the data that is being sent from all of the services to the audit-service where it is centrally processed.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/request/PolicyRequest.java
m78233: See the comment for OriginalRequest

results-service/pom.xml
m78233: Done

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/request/OriginalRequest.java
m78233: Your point about using the  @Generated annotation to declare that it does not need a unit test is right.  I'll make the necessary changes.

Have only identified one code smell that I needed to fix and did.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/response/UserResponse.java
m78233: Fixed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/response/UserResponse.java
m78233: Fixed

PR #148, opened 2020-07-07
PAL-287 first pass

nw1984:


Jenkinsfile
```diff
@@ -251,7 +251,32 @@ spec:
                               --set global.persistence.redisSlave.aws.volumeHandle=${VOLUME_HANDLE_REDIS_SLAVE} \
                               --namespace dev'
                         } else {
-                            sh "echo - no deploy"
+                            def GIT_BRANCH_NAME_LOWER = GIT_BRANCH_NAME.toLowerCase()
+                            sh 'palisade-login'
+                            //now extract the public IP addresses that this will be open on
+                            sh 'extract-addresses'
```
ac74475, line 257: I don't understand why this is required here?

nw1984: The extract-addresses shell script populates the values.yaml with the allowed ip addresses that are available for the helm chart. 
The cluster itself has an allowed ip address range.
This is also the allowed ip address range that the helm chart would allow access to.
In effect this is a layered approach to security.
If extract-addresses was not called, the helm chart (not the cluster) would allow all ip address range through.

PR #150, opened 2020-07-10
Pal 543 create k8s job to run example

developer01189998819991197253:


m78233 (dismissed):

data-service/src/main/java/uk/gov/gchq/palisade/service/data/config/ApplicationConfiguration.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Crown Copyright
+ * Copyright 2020 Crown Copyright
```
ac74475: revert, there are no other changes to this file

licenses/eclipse_public_license_1.0.html
```diff
@@ -15,20 +15,20 @@
     <link rel="shortcut icon" href="/eclipse.org-common/themes/solstice/public/images/favicon.ico"/>
```
ac74475: remove this licence change

resource-service/src/main/java/uk/gov/gchq/palisade/service/resource/service/StreamingResourceServiceProxy.java
```diff
@@ -171,11 +171,11 @@ public Boolean addResource(final LeafResource leafResource) {
     public void initPostConstruct() {
         // Add resources to persistence
         LOGGER.info("Prepopulating using resource builder: {}", resourceBuilder);
-        resourceBuilder.get().stream()
-                .peek(entry -> LOGGER.debug(entry.toString()))
+        resourceBuilder.get()
                 .forEach(entry -> {
                     Resource rootResource = entry.getKey();
                     LeafResource leafResource = entry.getValue();
+                    LOGGER.info("Persistence add for {} -> {}", rootResource.getId(), leafResource.getId());
```
ac74475: only really want this to be logged at debug level in a production system

values.yaml
```diff
@@ -65,6 +65,15 @@ global:
...
   deployment: codeRelease
 
+  # set environment variables for all palisade services pods by deployment type eg. spring profiles active
+  env:
+    _default:
+    - name: "SPRING_PROFILES_ACTIVE"
+      value: "k8s"
+    codeRelease:
+    - name: "SPRING_PROFILES_ACTIVE"
+      value: "k8s"
```
ac74475: why a _default and a code release?

ac74475 (changes requested):

resource-service/src/main/java/uk/gov/gchq/palisade/service/resource/service/StreamingResourceServiceProxy.java
dev930018: In all other services this is the current behaviour, changed for consistency.
Could have all changed to DEBUG but as we are changing some caching things, i think it's useful to keep at info for a bit

values.yaml
dev930018: Set different profiles depending upon deploy type. By default we want to run the jars with the k8s profile, on codeRelease we also want the k8s profile, but in example deployment we want to have example-prepopulation data loaded, so need the example-libs profile AND k8s

data-service/src/main/java/uk/gov/gchq/palisade/service/data/config/ApplicationConfiguration.java
developer01189998819991197253: removed

licenses/eclipse_public_license_1.0.html
developer01189998819991197253: removed

values.yaml
ac74475: delete code release one then

ac74475 (changes requested):
Just need to delete the 3 lines in the values.yaml for the codeRelease env variables so we just use the defaults. Given they are both the same

values.yaml
developer01189998819991197253: deleted

ac74475 (approved):

PR #154, opened 2020-07-17
PAL-697 change serviceClass to serviceName in RegisterRequestExceptionAuditRequest

dbz120896:
This resolves an issue where the audit service cannot de-serialise the service class.


pd104923 (dismissed):

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/SimplePalisadeService.java
```diff
@@ -179,7 +179,7 @@ private void auditRequestReceivedException(final RegisterDataRequest request, fi
         LOGGER.info("Auditing error with audit service");
         AuditRequest auditException = getAuditException(request)
                 .withException(ex)
-                .withServiceClass(serviceClass);
+                .withServiceName(serviceClass.getSimpleName());
         auditService.audit(auditException);
```
dev930018: I'd prefer to change the argument passed to this method too.
Then each of the get FutureXxx would be more like:
```
        }).exceptionally((Throwable ex) -> {
            auditRequestReceivedException(request, ex, "policy-service");
            throw new RegisterRequestException("Exception from policyService", ex);
        });
```

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/request/AuditRequestTest.java
```diff
@@ -91,7 +91,7 @@ public void RegisterRequestExceptionAuditRequestTest() {
                 .withResourceId("resource")
                 .withContext(new Context(Stream.of(new SimpleImmutableEntry<String, Class<?>>("a reason for access", String.class)).collect(toMap(SimpleImmutableEntry::getKey, SimpleImmutableEntry::getValue))))
                 .withException(new SecurityException("not allowed"))
-                .withServiceClass(UserService.class);
+                .withServiceName(UserService.class.getSimpleName());
```
dev930018: I'd also prefer to use just a static "user-service" string

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/SimplePalisadeService.java
dev930018: Since this is being sent out of the palisade-service to the audit-service, the general pattern on names is "something-service", this isn't too important internal to a service for eg logging but externally I think this consistency would be good (see SimpleConnectionDetail and resources/application.yaml web.client keys (values vary by environment))

dev930018 (changes requested):

dev930018:
Would also expect similar changes in audit-service classes

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/SimplePalisadeService.java
TRT-DarrenB: Would this be better to have `service.getName()`?

dbz120896: Would this be better to have `service.getName()`?

dbz120896: Would it be better to add `getType()` and `getServiceName()` default methods to the `Service` interface in common?

Something like:

```
/**
 * Returns the type of information managed by this service, e.g. "users".
 * If this method is not overridden, it default to the lowercase simple class name, e.g. "userservice".
 */
default public String getType() {
    return this.getClass().getSimpleName().toLowerCase();
}

/**
 * Returns the qualified name of this service, e.g. "palisade.service.users"
 */
default public String getName() {
    return "palisade.service." + getType();
} 
```
Just an idea.

dbz120896: Turns out that services do not have to implement the `uk.gov.gchq.palisade.service.Service` interface. With this being the case, there is no point in putting type and name methods onto it. The downside of this is that it is up external classes to create the name of the service when, for example, throwing an exception.

We'll stick to "something-service".

dbz120896:
I've added the static service names to the other services and now using that to pass to `SimplePalisadeService.auditRequestReceivedException` as per suggestions.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/SimplePalisadeService.java
```diff
@@ -50,6 +49,12 @@
  * registerDataRequest. </p>
  */
 public class SimplePalisadeService implements PalisadeService {
+
+    /**
+     * Service name
+     */
+    public static final String NAME = "simple-service";
```
dev930018: "palisade-service" ? maybe "simple-palisade-service"

dev930018 (changes requested):

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/SimplePalisadeService.java
ac74475: I think this should actually sit in the PalisadeService class and be inherited here such that it is actually named "palisade-service"

pd104923: Moved and given "palisade-service" value

dev930018 (dismissed):

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/PalisadeService.java
```diff
@@ -30,6 +30,11 @@
  */
 public interface PalisadeService extends Service {
 
+    /**
+     * Service name
+     */
+    String NAME = "palisade-service";
```
ac74475: Currently causing a SonarQube error
```suggestion
    public static final String NAME = "palisade-service";
```

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/request/AuditRequestTest.java
ac74475:
```suggestion
                .withServiceName(UserService.Name);
```

ac74475 (changes requested):
Should only need the image tag updating for the palisade service since none of the other services have materially changed

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/PalisadeService.java
pd104923: SonarQube doesn't like this being in an interface. Smell states that constants should be in a class or enum, will move it back to SimplePalisadeService and make it a static constant with the same value

ac74475: So SQ doesn't complain at the other interfaces where it is a static final constant.

pd104923: The other constants are in classes, not interfaces within the Palisade-service module.

palisade-service/src/test/java/uk/gov/gchq/palisade/service/palisade/request/AuditRequestTest.java
pd104923: Resolved

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/PalisadeService.java
pd104923: All service names have been put into an enum.

dev930018 (approved):

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
```diff
@@ -147,7 +146,7 @@ public void auditRegisterRequestUserException() {
                 .withResourceId(resource.getId())
                 .withContext(context)
                 .withException(exception)
-                .withServiceClass(UserService.class);
+                .withServiceName("USER_SERVICE");
```
dev930018: Add the same enum to audit-service too?

dev930018 (changes requested):

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
```diff
@@ -114,7 +114,7 @@ static Rules mockRules() {
                 .withResourceId(mockResource().getId())
                 .withContext(mockContext())
                 .withException(mockException())
-                .withServiceClass(Service.class);
+                .withServiceName("Service");
```
ac74475: We shouldn't ever have a registerRequestException with the service name being "Service". It should be split into 4 methods with the service coming from Palisade service, User service, Resource service and Policy service. These should also use the Enum class

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/LoggerAuditServiceTest.java
```diff
@@ -105,7 +104,7 @@ public void auditRegisterRequestException() {
                 .withResourceId(resource.getId())
                 .withContext(context)
                 .withException(exception)
-                .withServiceClass(Service.class);
+                .withServiceName("Service");
```
ac74475: Use one of the Enums, otherwise people might think that it is acceptable to put "service" in here.

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
ac74475: I agree with @dev930018 comments to add the serviceName enum into the audit service module

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
```diff
@@ -178,7 +177,7 @@ public void auditRegisterRequestResourceException() {
                 .withResourceId(resource.getId())
                 .withContext(context)
                 .withException(exception)
-                .withServiceClass(ResourceService.class);
+                .withServiceName("RESOURCE_SERVICE");
```
ac74475: use the enum

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
```diff
@@ -208,7 +207,7 @@ public void auditRegisterRequestOtherException() {
                 .withResourceId(resource.getId())
                 .withContext(context)
                 .withException(exception)
-                .withServiceClass(Service.class);
+                .withServiceName("Service");
```
ac74475: use an enum

ac74475 (changes requested):

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
dbz120896: `ServiceName` is in `palisade-service`. Should the enum be copied or the enum moved to common?

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/LoggerAuditServiceTest.java
dbz120896: `ServiceName` is in `palisade-service`. Should the enum be copied or the enum moved to common?

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
dbz120896: I'm not sure where the 4 methods come in, but that method has been changed.

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/LoggerAuditServiceTest.java
dbz120896: done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
dbz120896: done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
dbz120896: done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
dbz120896: done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
dbz120896: done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/ServiceName.java
```diff
@@ -0,0 +1,32 @@
...
+
+package uk.gov.gchq.palisade.service.audit.service;
+
+/**
+ * Enum providing type safety for service names
+ *
+ * @since 0.4.0
+ */
+public enum ServiceName {
+    TEST_SERVICE,
```
ac74475: This class should be in src rather than test. It also shouldn't contain "TEST_SERVICE"

audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditService.java
```diff
@@ -207,12 +207,12 @@ private static void onRegisterRequestException(final DefaultEventLoggingService
         stroomResource.setId(registerRequestExceptionAuditRequest.resourceId);
         authorise.getObjects().add(stroomResource);
         Outcome outcome = createOutcome(false);
-        if (registerRequestExceptionAuditRequest.serviceClass.getSimpleName().equalsIgnoreCase("UserService")) {
+        if ("USER_SERVICE".equalsIgnoreCase(registerRequestExceptionAuditRequest.serviceName)) {
```
ac74475:
```suggestion
        if (ServiceName.USER_SERVICE.name().equalsIgnoreCase(registerRequestExceptionAuditRequest.serviceName)) {
```

audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditService.java
```diff
@@ -207,12 +207,12 @@ private static void onRegisterRequestException(final DefaultEventLoggingService
...
             LOGGER.debug("onRegisterRequestException  registerRequestExceptionAuditRequest is UserService");
             exceptionEventDetail.setTypeId(REGISTER_REQUEST_EXCEPTION_USER_TYPE_ID);
             exceptionEventDetail.setDescription(REGISTER_REQUEST_EXCEPTION_USER_DESCRIPTION);
             outcome.setDescription(REGISTER_REQUEST_EXCEPTION_USER_OUTCOME_DESCRIPTION);
-        } else if (registerRequestExceptionAuditRequest.serviceClass.getSimpleName().equalsIgnoreCase("ResourceService")) {
+        } else if ("RESOURCE_SERVICE".equalsIgnoreCase(registerRequestExceptionAuditRequest.serviceName)) {
```
ac74475:
```suggestion
        } else if (ServiceName.RESOURCE_SERVICE.name().equalsIgnoreCase(registerRequestExceptionAuditRequest.serviceName)) {
```

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
```diff
@@ -114,7 +114,7 @@ static Rules mockRules() {
                 .withResourceId(mockResource().getId())
                 .withContext(mockContext())
                 .withException(mockException())
-                .withServiceClass(Service.class);
+                .withServiceName(ServiceName.USER_SERVICE.name());
     }
```
ac74475, line 118: Given this is a common utils class for testing, I think we need to make it clear that this method is registerRequestExceptionAuditRequestFromUserService. And then if we need the exception to have come from a different service then we can have another similarly named method for that.

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
```diff
@@ -208,7 +208,7 @@ public void auditRegisterRequestOtherException() {
                 .withResourceId(resource.getId())
                 .withContext(context)
                 .withException(exception)
-                .withServiceClass(Service.class);
+                .withServiceName(ServiceName.TEST_SERVICE.name());
```
ac74475:
```suggestion
                .withServiceName(ServiceName.POLICY_SERVICE.name());
```

ac74475 (changes requested):

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/ServiceName.java
dbz120896: I don't think this should be in `src/main/java` as it will be an unused class in the jar. This class is only used at test scope.

dbz120896: Done

audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditService.java
dbz120896: Done

audit-service/src/main/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditService.java
dbz120896: Done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/AuditServiceTestCommon.java
dbz120896: Done

audit-service/src/test/java/uk/gov/gchq/palisade/service/audit/service/StroomAuditServiceTest.java
dbz120896: Done

ac74475 (approved):

dev930018 (approved):

PR #155, opened 2020-07-28
PAL-695 Jenkins Build Times

dbz120896:
Just added timestamps and the sonarqube section to name the analysis by the branch name if it's a PR.


developer01189998819991197253 (changes requested):
Whats with all the whitespace changes? 
The SonarQube stage needs to be replicated in other repositories too

Jenkinsfile
```diff
@@ -26,274 +26,280 @@
...
+        imagePullPolicy: IfNotPresent
+        command:
+        - sleep
+        args:
+        - 99d
+        env:
+          - name: DOCKER_HOST
+            value: tcp://localhost:2375
+        resources:
+          requests:
```
ac74475:
```suggestion
          requests:
            cpu: "2"
```

dbz120896: Aha! That's where :)

dbz120896: I take it this would be in all the other repositories as well?

dbz120896:
> Whats with all the whitespace changes?
> The SonarQube stage needs to be replicated in other repositories too

I've added the timestamps plugin. Everything inside that is now indented. Maybe that did it.

pd104923 (approved):

developer01189998819991197253 (approved):

PR #156, opened 2020-07-28
Pal 705 fix perf test errors

pd104923:


pom.xml
```diff
@@ -21,7 +21,7 @@
     <parent>
         <artifactId>spring-boot-starter-parent</artifactId>
         <groupId>org.springframework.boot</groupId>
-        <version>2.2.0.RELEASE</version>
+        <version>2.3.1.RELEASE</version>
```
developer01189998819991197253: why the huge jump in spring versions?

developer01189998819991197253 (changes requested):

pom.xml
pd104923: This was added as a test in order to get the trials working. I don't think it's actually relevant to the problem so will try and remove it (and the other pom changes) and run the trials locally.

pd104923: Removing this does not affect the performance trials. I have therefore removed it (and the palisade and resource pom updates)

dbz120896 (approved):

developer01189998819991197253 (approved):

dev-958 (approved):

ac74475 (approved):

PR #157, opened 2020-07-30
Pal 645 updated hadolint image

pd104923:


Jenkinsfile
```diff
@@ -75,7 +75,7 @@ spec:
         ephemeral-storage: "8Gi"
   
   - name: hadolint
-    image: hadolint/hadolint:latest-debian@sha256:15016b18964c5e623bd2677661a0be3c00ffa85ef3129b11acf814000872861e
+    image: hadolint/hadolint:v1.18.0-6-ga0d655d-alpine@sha256:e0f960b5acf09ccbf092ec1e8f250cd6b5c9a586a0e9783c53618d76590b6aec
```
kg981167224, line 78: More for my understanding Is there a reason we don't just pull the latest?

kg981167224 (approved):

Jenkinsfile
pd104923: This is done for security reasons. @ac74475 should be able to give you more details on the reasoning.

dev930018 (approved):

ac74475 (approved):

PR #159, opened 2020-07-30
Pal 564 k8s performance tests

dev930018:


resource-service/src/main/java/uk/gov/gchq/palisade/service/resource/service/StreamingResourceServiceProxy.java
```diff
@@ -175,7 +175,7 @@ public void initPostConstruct() {
                 .forEach(entry -> {
                     Resource rootResource = entry.getKey();
                     LeafResource leafResource = entry.getValue();
-                    LOGGER.debug("Persistence add for {} -> {}", rootResource.getId(), leafResource.getId());
+                    LOGGER.info("Persistence add for {} -> {}", rootResource.getId(), leafResource.getId());
```
pd104923, line 178: Is there a reason that this has been changed from `debug`? Surely this would add a lot of extra logging into the resource-service logs (all be-it at the start of the logs).

pd104923 (changes requested):
Just one little question.

resource-service/src/main/java/uk/gov/gchq/palisade/service/resource/service/StreamingResourceServiceProxy.java
dev930018: I was doing this before to keep in-line with the style in user/policy (log at info all things pre-poulated), but we currently do not have a check that this method has completed before the resource-service is declared as UP and healthy. A standard run of the perf trials requires a `kubectl logs -f resource-service-xxxxx` and waiting until all resources have been added.
If this were removed, we'd either have to wait some arbitrary amount of time and guess, or run the whole service at debug.

I think generally worth keeping, only causes as much mess as is in the prepop yaml (i.e. a prepop add for `/resources/many/dir/` will not prepop add `/resources/many/dir/file00999`

pd104923: Yeah, it's difficult to know when to start the process as we don't specifically know when the prepop has finished. 
Happy for this to stay as it is.

pd104923 (dismissed):

developer01189998819991197253 (approved):

pd104923 (approved):

PR #161, opened 2020-08-05
Pal 728 added check to create topic script

pd104923:


dbz120896 (dismissed):
Looks :+1:. There are a couple of bash constructs in there that I'm going to look up later.

create-kafka-queues/create-topics.sh
```diff
@@ -54,9 +54,17 @@ else
 
   #Search for all environmental variables starting with the word: TOPIC
   for topic in "${!TOPIC@}"; do
```
dbz120896, line 56: I get what the `!` does, but what does the `@` do? From googling, all I get is arrays.

pd104923: This is searching for all env variables that have "TOPIC" at the start. This will create an array of these values that we then use in the rest of the script. These values are created through the `create-kafka-queues` values.yaml file.

m78233 (dismissed):

developer01189998819991197253:
@pd104923 why are there so many dependency tree updated in this PR if there are no changes to any versions?

pd104923:
> @pd104923 why are there so many dependency tree updated in this PR if there are no changes to any versions?

This is due to the parent pom having the Spring Boot Parent updated to 2.3.1 which had not been updated in the other dependency trees. I had performed a clean install locally which is why all these files have been updated.

ac74475 (dismissed):

developer01189998819991197253 (changes requested):
when deploying on WSL it fails, when doing a kubectl logs create-kafka-pod it returns:
/bin/create-topics.sh: line 22: $'\r': command not found
/bin/create-topics.sh: line 28: syntax error near unexpected token $'\r'
/bin/create-topics.sh: line 28: write_to_kafka () {

Youll need to convert the line endings appropriatley

dev-958 (approved):

developer01189998819991197253 (approved):

PR #164, opened 2020-08-17
PAL-717 Added binary switch for redis master/worker or clustered

dbz120896:
Added a binary switch that can be used to select which Redis install to use. This does not effect the installation. That is still controlled by two other switches. This enables installation of Palisade without installing any Redis types, but selecting which existing one to use.

Could devs have a look at this. I'm sure there is something missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,16 +117,19 @@ Some more important arguments are as follows:
 | global.kafka.install                    | Install Kafka and Zookeeper, **default=true**
 | global.redis.install                    | Install Redis, **default=true**
 | global.redis-cluster.install            | Install Redis-cluster, **default=false**
+| global.redisClusterEnabled              | Set to true to use Redis cluster. Useful if redis is already installed. **default=false**
 
 #### Redis vs Redis-Cluster
+
 The key difference is scalability, write-points, sharding and partitioning.
-* Redis will support a single (master) write-point with many replicated (slave) read-points.
+* Redis will support a single (master) write-point with many replicated (worker) read-points.
 * Redis-cluster will support sharding across the keyspace such that keys are uniquely mapped to one of many partitions, each partition with a single write-point and many read-points as above.
 
 Recommended reading: [Amazon AWS documentation, Redis with Cluster Mode enabled](https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Replication.Redis-RedisCluster.html)
 Redis is simpler, but will bottle-neck on write-requests to the single master node, eventually.
 
 #### Kubernetes Dashboard
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -117,16 +117,19 @@ Some more important arguments are as follows:
 | global.kafka.install                    | Install Kafka and Zookeeper, **default=true**
 | global.redis.install                    | Install Redis, **default=true**
```
kg981167224, line 130: Could global.redis.install &  global.redis-cluster.install be removed now?

kg981167224 (commented):
jenkinsfile has  global.redis-cluster.install references, which are currently commented out; however I think Alex will be re-enabling this very soon, so is it worth seeing if the jenkinsfile can be changed or let alex know?

README.md
dbz120896: Those cannot be removed. The new switch is simply to select which should be used. The install flags still need to be there. It's annoying that the condition in Chart.yaml cannot support not (`!flag`). I scratched my head for a little bit about this one. I didn't want to play around with the Jenkinsfile yet.

pd104923 (dismissed):

templates/NOTES.txt
```diff
@@ -24,3 +24,13 @@ You have installed {{ .Chart.Name }} as release {{ .Release.Name }} into namespa
...
 {{- end }}
+
+{{- if $.Values.global.redisClusterEnabled }}
+    {{- $clusterNodes := include "palisade.redis-cluster.headlessNodes" . | trimAll "," }}
+    {{- println }}Redis is clustered using nodes: {{ $clusterNodes }}
+{{- else }}
+    {{- $masterFullname := printf "%s-master" (include "palisade.redis.fullname" .) }}
+    {{- $masterPort := int .Values.global.redis.master.service.port }}
+    {{- println }}Redis is master/worker with master at: {{ $masterFullname }}:{{ $masterPort }}
+{{- end }}
```
dev930018, line 35: I'd like to see a warning here for the specific case when redis(cluster) is set to install but does not match the flag:
```
if (redisClusterEnabled and redis.install) or (not redisClusterEnabled and redisCluster.install)
println Redis install type does not match redis deployment type, are you sure you know what you are doing?
```

dev930018 (changes requested):

templates/NOTES.txt
dbz120896: I found that you can now get it to fail if you have an install/use mismatch, which I think is better. If no installation is selected, then no error will be thrown.

dev930018 (dismissed):

ac74475 (changes requested):
Given the limitations on the charts dependancy conditionals, I think the best option is to work with just the original two Redis install parameters and have a check that will fail the deployment if both are set too true. Rather than adding in another parameter.

ac74475 (approved):

PR #165, opened 2020-08-18
Pal 718 global flag for statefulset pv/pvc(s)

m78233:
Added a global flag that can be overwritten that will exempt the pv/pvc from being re-built.  Made changes to the helm charts and the readme file.  Did not change the Jenkins file as it did not seem to be relevant.


m78233:
Branch name is not acceptable for SonarQube.  Have changed the branch name and re-submitted the pull request

PR #168, opened 2020-08-24
Pal 743 aws dynamic storage

dev-958:
Currently deploys feature branches on the cluster but there is more work to be done, detailed in PAL-710, to get develop branch deployment working correctly.


values.yaml
```diff
@@ -64,7 +64,7 @@ global:
       nodes: 6
 
   # hosting: [local | aws | metal] the target environment to deploy to
-  hosting: local
+  hosting: aws
```
dev930018: What do we want the default to be?
At the moment, a `helm install palisade .` works locally, with this we would have to be slightly more explicit in what is probably the more-common use-case at the moment (and in general, any aws deploys are probably automated so whether one is the default or not doesn't effect time typing commands)

values.yaml
```diff
@@ -103,11 +103,19 @@ global:
...
         local: hostpath
         aws: efs-sc
+      auto:
+        _default: false
+        local: false
+        aws: false
+      mode:
+        _default: ReadWriteOnce
+        local: ReadWriteMany
+        aws: ReadWriteMany
```
dev930018, line 113: These only need to be ReadOnlyMany, unless that still isn't properly supported by the aws drivers?
I'd expect _default and local to be ReadOnlyMany if possible and aws ReadWriteMany if need be

values.yaml
```diff
@@ -128,6 +136,14 @@ global:
...
           local: hostpath
           aws: efs-sc
+        auto:
+          _default: false
+          local: false
+          aws: false
+        mode:
+          _default: ReadWriteOnce
+          local: ReadWriteMany
+          aws: ReadWriteMany
```
dev930018, line 146: Ditto here, should only need to be ReadOnlyMany

values.yaml
```diff
@@ -181,9 +222,17 @@ global:
...
       name: redis-persistence-slave
-      storage: 4Gi
+      auto:
+        _default: true
+        local: true
+        aws: true
+      mode:
+        _default: ReadWriteOnce
+        local: ReadWriteOnce
+        aws: ReadWriteOnce
```
dev930018, line 234: I presume the statefulset redis/slave/{n} mount path means this would only ever be ReadWriteOnce?

values.yaml
dev-958: I'll switch that back to local, forgot it.

values.yaml
dev-958: In aws it's not supported under EFS. I've made it ReadWriteMany for the local profile because it's become clear that deploying the application on the infrastructure is going to be inflexible in cases where you have heterogeneous storage typologies and may require dynamic provisioning. Going to raise a ticket to provide a rest endpoint for deploying jars onto the storage.

values.yaml
dev-958: as above.

values.yaml
dev-958: Exactly.

dev930018 (dismissed):

developer01189998819991197253 (approved):

pd104923 (approved):

dev930018 (approved):

Jenkinsfile
```diff
@@ -242,72 +242,77 @@ spec:
...
-//                 }
-//             }
-//         }
+
+         stage('Helm deploy') {
+             dir('Palisade-services') {
+                 container('maven') {
+                     configFileProvider([configFile(fileId: "${env.CONFIG_FILE}", variable: 'MAVEN_SETTINGS')]) {
+                         if (("${env.BRANCH_NAME}" == "develop") ||
+                                 ("${env.BRANCH_NAME}" == "master")) {
```
ac74475, line 251:
```suggestion
                                 ("${env.BRANCH_NAME}" == "main")) {
```

ac74475 (approved):

PR #180, opened 2020-09-16
Pal 815 update create kafka queue

pd104923:


dbz120896 (dismissed):

charts/create-kafka-queues/values.yaml
```diff
@@ -27,19 +27,19 @@ kafkaCreateTopics:
   - name: resource
     replicationFactor: 1
     partitions: 1
-  - name: policy
+  - name: rule
     replicationFactor: 1
     partitions: 1
-  - name: query-scope
+  - name: filtered-resource
```
dev930018: Approved, but I feel there is a stronger case for calling this `masked-resource`
The queues are named after the items put on them, so user has messages with a user etc.
Everything after policy-service applying resource-level rules is 'filtered', in that we have totally removed some elements of the stream, so the name `filtered-resource` does not uniquely apply to this step.
The services in general add something to a message, so have a downstream queue named after its upstream service in general (user-service -> user queue, etc.)
What has happened after the attribute-masking-service is that resource attributes have been masked, so the queue holds a collection of `masked-resource`s (which are also filtered-and-with-rules, with-resources, with-users)

dev930018 (dismissed):

charts/create-kafka-queues/values.yaml
pd104923: This has been updated to `masked-resource` and the offset topic has been updated to `masked-resource-offset`

dev930018 (approved):

ac74475 (approved):

PR #184, opened 2020-09-25
Pal 544 k8s end to end

pd104923:


charts/policy-service/values.yaml
```diff
@@ -20,7 +20,7 @@ service:
 
 image:
   name: policy-service
-  tag: SNAPSHOT-026cffd
```
m78233, line 23: More a question than anything else.  This tag is only being changed in this one service and yet this same tag appears in the other services with the older value.  I would expect that this will be the same number for all of them.  Can I assume that the tags for all the services get updated at some other point?

m78233 (commented):
A question on one of the changes.

charts/policy-service/values.yaml
pd104923: The other services do not need to have their tag updated as there are no changes to them. The tags should only be updated on the service(s) where there are changes.

pd104923: Having said that there are no real changes to the service but the `pi` profile was used to make sure the `deployment.yaml` changes were used when deploying the service.

m78233 (approved):

dev930018 (approved):

ac74475 (approved):

PR #213, opened 2020-11-27
Pal 926 build without docker pull

kg981167224:
Just a small part of PAL-926, to use a standard Java 11 library for all containers.

The rest of the change is a separate review for the scripts.


create-kafka-queues/Dockerfile
```diff
@@ -12,17 +12,7 @@
...
-RUN apk update
-#Search for a particular version of dependancies
-#RUN apk search -v 'vim*'
-#RUN apk search -v 'curl*'
-#RUN apk search -v 'bash*'
-
-RUN apk --no-cache add vim=8.1.1365-r0
-RUN apk --no-cache add curl=7.64.0-r4
-RUN apk --no-cache add bash=4.4.19-r1
+FROM openjdk:11@sha256:736dd07fc5cb53060a49d38c86b9e673974510c5a18afc40b1c936475f59aee7
```
developer01189998819991197253, line 15: Does openjdk11 include curl, bash and vim?

developer01189998819991197253 (commented):
Question about curl bash and vim apk adds

create-kafka-queues/Dockerfile
kg981167224: The openjdk 11 uses the full Debian distribution, so looks to include everything we need already. Whereas the alpine variant is approx 1/8th the size but missing a few bits.

We may re-visit Alpine later to see if we can reduce image size a little, but for now we are going to use a standard base image for all containers.

m78233 (approved):
looks good

create-kafka-queues/Dockerfile
dev930018: The lifecycle of this container is in the order of seconds, and only on a fresh install.
For all others i could see why we'd standardize to a fat feature-full image, but it seems rather extreme for this job

developer01189998819991197253 (approved):

dev930018 (approved):

dev-958 (approved):

PR #215, opened 2020-11-27
Pal 869 user service auditing

developer01189998819991197253:


developer01189998819991197253:
After PR has been approved I will squash and recommit to remove unsigned commits

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+--->
+
+# <img src="../logos/logo.svg" width="180">
+
 # User Service
 
-<span style="color:red">**Note:** As noted in the [documentation root](../README.md) Palisade is in an early stage of development, therefore the precise APIs contained in these service modules are likely to adapt as the project matures.</span>
+The User Service is responsible for providing the other Palisade components with knowledge of the users of the system. 
+Note that we are avoiding having a separate notion of "Palisade" users that are distinct from the normal user accounts of the system. There is no such thing as a "Palisade" user.  
+One of the purposes of this service is to allow Palisade to adopt the whatever notion of *user* the host environment has. For example, this may from a central directory service such as LDAP,
```
kg981167224:
```suggestion
One of the purposes of this service is to allow Palisade to adopt whatever notion of *user* the host environment has. For example, this may from a central directory service such as LDAP, 
```

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/CaffeineUserCachingTest.java
```diff
@@ -65,91 +62,79 @@ private void forceCleanUp() {
...
 
         // When
-        User getUser = userService.getUser(request).join();
+        User getUser = userService.getUser(user.getUserId().getId());
         // Then
         assertThat(getUser).isEqualTo(user);
     }
 
     @Test
     void testNonExistentUserRetrieveFails() {
```
kg981167224, line 78: Is this test reliable? Since there is a sleep for 1s and TTL is 1s isn't there a chance the cache doesn't get cleaned up, especially if the sleep gets an interrupt.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/CaffeineUserCachingTest.java
```diff
@@ -65,91 +62,79 @@ private void forceCleanUp() {
...
 
     @Test
     void testTtl() throws InterruptedException {
         // Given - a user was added a long time ago (ttl set to 1s in application.yaml)
         User user = new User().userId("ttl-test-user").addAuths(Collections.singleton("authorisation")).addRoles(Collections.singleton("role"));
-        UserRequest request = UserRequest.Builder.create().withUserId(user.getUserId().getId()).withResourceId("test/resource").withContext(new Context().purpose("test"));
         userService.addUser(user);
 
-        TimeUnit.MILLISECONDS.sleep(1000);
+        TimeUnit.SECONDS.sleep(1);
```
kg981167224, line 130: See earlier question on sleep.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
```diff
@@ -0,0 +1,180 @@
...
+        // Then - it is no longer found, it has been evicted
+        // ie. throw NoSuchUserIdException
+        assertThat(noSuchUserId.getMessage()).isEqualTo("No userId matching ttl-test-user found in cache");
+    }
+
+    public static class RedisInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+        private static final int REDIS_PORT = 6379;
+
+        static final GenericContainer<?> REDIS = new GenericContainer<>("redis:6-alpine")
```
kg981167224, line 160: Is this going to pull this image every time? Or will it use a local cache. Maybe worth testing to ensure we aren't going to hit any rate issues. Especially when run in pipeline

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
```diff
@@ -0,0 +1,180 @@
...
+        // Given - a user was added a long time ago (ttl set to 1s in application.yaml)
+        User user = new User().userId("ttl-test-user").addAuths(Collections.singleton("authorisation")).addRoles(Collections.singleton("role"));
+        UserRequest request = UserRequest.Builder.create().withUserId(user.getUserId().getId()).withResourceId("test/resource").withContext(new Context().purpose("purpose"));
+        cacheProxy.addUser(user);
+
+        TimeUnit.MILLISECONDS.sleep(1000);
+
+        // When - we try to access stale cache data
+        Exception noSuchUserId = assertThrows(NoSuchUserIdException.class,
+                () -> cacheProxy.getUser(request.userId), "testMaxSizeTest should throw noSuchIdException"
```
kg981167224:
```suggestion
                () -> cacheProxy.getUser(request.userId), "testTTL should throw noSuchIdException"
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/UserApplication.java
```diff
@@ -88,30 +86,27 @@ public UserApplication(
      * @param args command-line arguments passed to the application
      */
     public static void main(final String[] args) {
-        LOGGER.debug("UserApplication started with: {} {} {}", UserApplication.class, "main", args);
+        LOGGER.debug("UserApplication started with: {}", (Object) args);
```
kg981167224, line 89: Should this be Object[] since you want to call debug(String var1, Object... var2)

Although this change gives a warning, and I think ultimately it will work anyway.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/model/AuditMessage.java
```diff
@@ -49,28 +49,23 @@
     protected static final ObjectMapper MAPPER = new ObjectMapper();
 
     @JsonProperty("userId")
-    protected final String userId; //Unique identifier for the user.
+    protected final String userId; // Unique identifier for the user.
 
     @JsonProperty("resourceId")
-    protected final String resourceId;  //Resource that that is being asked to access.
+    protected final String resourceId;  // Resource that that is being asked to access.
```
kg981167224, line 54: Remove blank lines here to match rest of removed lines?

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,101 @@
...
+                    .all()
+                    .get(1, TimeUnit.SECONDS);
+
+            ConsumerGroupDescription consumerGroupDescription = groupDescriptionMap.get(this.groupId);
+
+            LOGGER.debug("Kafka consumer group ({}) state: {}", groupId, consumerGroupDescription.state());
+
+            if (consumerGroupDescription.state() == ConsumerGroupState.STABLE) {
+                return consumerGroupDescription.members().stream()
+                        .noneMatch(member -> (member.assignment() == null || member.assignment().topicPartitions().isEmpty()));
```
kg981167224: Would a LOGGER.error be good here if we fail and return false?

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
```diff
@@ -108,7 +108,7 @@ private SerDesConfig() {
     /**
      * Kafka value serialiser for downstream messages going out as output
      *
-     * @return an appropriate value serialiser for the topic's message content (AttributeMaskingResponse)
+     * @return an appropriate value serialiser for the topic's message content
```
kg981167224:
```suggestion
     * @return an appropriate value serializer for the topic's message content
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
```diff
@@ -120,6 +120,15 @@ private SerDesConfig() {
         };
     }
 
+    /**
+     * Kafka value serialiser for downstream messages going out as output
+     *
+     * @return an appropriate value serialiser for the topic's message content
```
kg981167224, line 126:
```suggestion
     * @return an appropriate value serializer for the topic's message content
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
```diff
@@ -120,6 +120,15 @@ private SerDesConfig() {
...
+     */
+    public static Serializer<byte[]> passthroughValueSerializer() {
+        return (String ignored, byte[] bytes) -> bytes;
+    }
+
     /**
      * Kafka key serialiser for downstream messages going out as errors
```
kg981167224, line 133:
```suggestion
     * Kafka key serializer for downstream messages going out as errors
```

kg981167224 (dismissed):
Nothing urgent so approved with comments.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/CaffeineUserCachingTest.java
developer01189998819991197253: the testNonExistentUserRetrieveFails test never adds a user to the cache, so gets a user that never exists and then throws the exception

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/CaffeineUserCachingTest.java
developer01189998819991197253: The test has been reliable so far, but also below runs the forceCleanUp() method that uses the cacheManager to clean up the cache manually if the sleep gets interrupted etc

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
developer01189998819991197253: it should use locally cached images

user-service/src/main/java/uk/gov/gchq/palisade/service/user/UserApplication.java
developer01189998819991197253: It should work fine this way, and matches other services for consistency also

user-service/src/main/java/uk/gov/gchq/palisade/service/user/model/AuditMessage.java
developer01189998819991197253: removed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
developer01189998819991197253: fixed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
developer01189998819991197253: fixed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/SerDesConfig.java
developer01189998819991197253: fixed

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
dev930018: Also, any rate issues would only be against redis's pull allowance, not ours, no?

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
dev930018: In the case you return `false`, the service will soon be shut-down by k8s.
Any loggers here would get called periodically (every 20s or so?).
I think a `LOGGER.error("Failed to find kafka topic-partition assignments")` would be reasonable in this case.
Then a `kubectl logs user-service-0 -p` would reveal that there was a kafka problem quite quickly.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
kg981167224: I think any test container pull will do the request to DH logging our IP address with docker hub, which will come off our allocation. TBH If it uses local cache won't be a problem.

kg981167224 (dismissed):

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
developer01189998819991197253: updated to add a log message

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+    "userId": {
+      "id": "test-user-id"
+    },
+    "roles": [
+      "roles"
+    ],
+    "auths": [
+      "auths"
+    ],
+    "class": "uk.gov.gchq.palisade.User"
```
dbz120896, line 98: Should not be returning class. This is exposing the service internal model, which is subject to change.

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+and then this is sent via kafka, to the Resource Service for further processing. 
+If the User does not exist in the cache, then a `NoSuchUserIdException` is thrown within the Service, which is caught, and packaged in the `AuditableUserReponse`, alongside the original `UserRequest` and an `AuditErrorMessage`, 
+this is then sent to the Audit Service and logged appropriatly.
 
-## API Usage
+## REST Interface
 
-Currently the `UserService` supports adding and retrieving of user details via the two following API calls.
+The application exposes two REST endpoints used for debugging or mocking kafka entrypoints:
+* `POST api/user`
```
dbz120896, line 54: We are not posting a user, but a request. The endpoint should be `api/users/requests` to keep in line with RESTful semantics.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
dbz120896: When run in Jenkins, this will indeed pull the container every time. We need to do the same as we have planned for Maven, and have a EFS volume which is shared.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
```diff
@@ -0,0 +1,195 @@
...
+@ContextConfiguration(initializers = {RedisInitializer.class})
+@Import(RedisAutoConfiguration.class)
+@ActiveProfiles({"redis"})
+class UserServiceAsyncProxyErrorTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserServiceAsyncProxyErrorTest.class);
+
+
+    private static final Function<Integer, String> REQUEST_FACTORY_JSON = i ->
+            String.format("{\"userId\":\"test-user-id\",\"resourceId\":\"/test/resourceId\",\"context\":{\"class\":\"uk.gov.gchq.palisade.Context\",\"contents\":{\"purpose\":\"purpose\"}}}", i, i);
```
dbz120896: You could do this:
```
String.format("{ 'userId' : 'test-user-id' }").replace("'", "\"");
```
Gets rid of all the unsightly escaping making it a bit easier to read, but is down to preference.

dbz120896 (changes requested):
Only real changes are the class in the response and the URL for the POST endpoint.

user-service/README.md
developer01189998819991197253: This would require a change to the other services also, as they follow the same convention of /api/*service e.g /api/mask
if its worth doing, we can create a ticket on the backlog

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
developer01189998819991197253: We could, but as we show an example of the pretty json in the README, there is no real gain in showing a pretty version here too

user-service/README.md
developer01189998819991197253: the rest post is an internal api that is only meant for debugging. Inside a cluster these endpoints wont be visible

dbz120896: Ok. We can let this go then.

user-service/README.md
dbz120896: I think it is worth doing, but is a low priority. An issue in the backlog would be best I think.

dbz120896 (dismissed):

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+--->
+
+# <img src="../logos/logo.svg" width="180">
+
 # User Service
 
-<span style="color:red">**Note:** As noted in the [documentation root](../README.md) Palisade is in an early stage of development, therefore the precise APIs contained in these service modules are likely to adapt as the project matures.</span>
+The User Service is responsible for providing the other Palisade components with knowledge of the users of the system. 
+Note that we are avoiding having a separate notion of "Palisade" users that are distinct from the normal user accounts of the system. There is no such thing as a "Palisade" user.  
+One of the purposes of this service is to allow Palisade to adopt whatever notion of *user* the host environment has. For example, this may from a central directory service such as LDAP,
```
dev930018:
```suggestion
One of the purposes of this service is to allow Palisade to adopt whatever notion of *user* the host environment has. For example, this may come from a central directory service such as LDAP, 
```

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+| userId          | userId           | userId            | UserResponse          |  
+| resourceId      | resourceId       | resourceId        | AuditErrorMessage     |
+| context         | context          | context           |                       |
+|                 | User             | exception         |                       | 
+|                 |                  | serverMetadata    |                       |
+  
+(fields marked with * are acquired from headers metadata)
 
-The user service separates this concern from the rest of the system. Other components use this service's API to request user details. Some deployments may also allow Palisade to add users to the system, hence the presence of the `addUser()` method in the `UserService` interface.
+The service accepts a `UserRequest` and a token, then checks that the User is in the cache; the technology of which is chosen in the User Services application.yaml and Spring profiles.
```
dev930018:
```suggestion
The service accepts a `UserRequest` and a token, then checks if the User is in the cache; the technology of which is chosen in the User Services application.yaml and Spring profiles.
```

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+If the User exists in the cache, then it, alongside any attributes are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors,
```
dev930018:
```suggestion
If the User exists in the cache, then it, alongside any attributes are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors, 
```
```suggestion
If the User exists in the cache, then it, alongside any additional attributes, are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors.
```

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+| context         | context          | context           |                       |
+|                 | User             | exception         |                       | 
+|                 |                  | serverMetadata    |                       |
+  
+(fields marked with * are acquired from headers metadata)
 
-The user service separates this concern from the rest of the system. Other components use this service's API to request user details. Some deployments may also allow Palisade to add users to the system, hence the presence of the `addUser()` method in the `UserService` interface.
+The service accepts a `UserRequest` and a token, then checks that the User is in the cache; the technology of which is chosen in the User Services application.yaml and Spring profiles.
+If the User exists in the cache, then it, alongside any attributes are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors, 
+and then this is sent via kafka, to the Resource Service for further processing.
```
dev930018:
```suggestion
Then, this is sent via Kafka to the [Resource Service](../resource-service) for further processing. 
```

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+|                 |                  | serverMetadata    |                       |
+  
+(fields marked with * are acquired from headers metadata)
 
-The user service separates this concern from the rest of the system. Other components use this service's API to request user details. Some deployments may also allow Palisade to add users to the system, hence the presence of the `addUser()` method in the `UserService` interface.
+The service accepts a `UserRequest` and a token, then checks that the User is in the cache; the technology of which is chosen in the User Services application.yaml and Spring profiles.
+If the User exists in the cache, then it, alongside any attributes are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors, 
+and then this is sent via kafka, to the Resource Service for further processing. 
+If the User does not exist in the cache, then a `NoSuchUserIdException` is thrown within the Service, which is caught, and packaged in the `AuditableUserReponse`, alongside the original `UserRequest` and an `AuditErrorMessage`,
```
dev930018: If it isn't found in the cache, it calls the real service, which should throw a `NoSuchUserIdException`. Only our 'trivial' implementation, the NullUserService, always throws this exception.

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
+|                 |                  | serverMetadata    |                       |
+  
+(fields marked with * are acquired from headers metadata)
 
-The user service separates this concern from the rest of the system. Other components use this service's API to request user details. Some deployments may also allow Palisade to add users to the system, hence the presence of the `addUser()` method in the `UserService` interface.
+The service accepts a `UserRequest` and a token, then checks that the User is in the cache; the technology of which is chosen in the User Services application.yaml and Spring profiles.
+If the User exists in the cache, then it, alongside any attributes are retrieved, and used to create a `UserReponse` object which is then packaged into a `AuditableUserResponse` object with no errors, 
+and then this is sent via kafka, to the Resource Service for further processing. 
+If the User does not exist in the cache, then a `NoSuchUserIdException` is thrown within the Service, which is caught, and packaged in the `AuditableUserReponse`, alongside the original `UserRequest` and an `AuditErrorMessage`, 
+this is then sent to the Audit Service and logged appropriatly.
```
dev930018:
```suggestion
this is then sent to the [Audit Service](../audit-service) and audited appropriately.
```

user-service/README.md
dev930018: I like the usage we currently have as it is meant for pushing items onto kafka queues. In this way, the REST endpoint reflects the name of the topic.

user-service/README.md
```diff
@@ -1,26 +1,104 @@
...
 
-## API Usage
+## REST Interface
 
-Currently the `UserService` supports adding and retrieving of user details via the two following API calls.
+The application exposes two REST endpoints used for debugging or mocking kafka entrypoints:
```
dev930018, line 53: "exposes two" but have only described one (`api/user/multi` is missing)

user-service/README.md
dev930018: If anything, you need this class as it may be a `uk.gov.gchq.palisade.example.ExampleUser` or some other richer object with extra fields

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Test
+    void testTtl() throws InterruptedException {
+        // Given - a user was added a long time ago (ttl set to 1s in application.yaml)
+        User user = new User().userId("ttl-test-user").addAuths(Collections.singleton("authorisation")).addRoles(Collections.singleton("role"));
+        UserRequest request = UserRequest.Builder.create().withUserId(user.getUserId().getId()).withResourceId("test/resource").withContext(new Context().purpose("purpose"));
+        cacheProxy.addUser(user);
+
+        TimeUnit.MILLISECONDS.sleep(1000);
```
dev930018:
```suggestion
        TimeUnit.SECONDS.sleep(1);
```

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaInitializer.java
```diff
@@ -0,0 +1,155 @@
...
+    @Override
+    public void initialize(final ConfigurableApplicationContext configurableApplicationContext) {
+        configurableApplicationContext.getEnvironment().setActiveProfiles("akka-test", "caffeine", "pre-population", "debug");
+        KAFKA_CONTAINER.addEnv("KAFKA_AUTO_CREATE_TOPICS_ENABLE", "false");
+        KAFKA_CONTAINER.addEnv("KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR", "1");
+        KAFKA_CONTAINER.start();
+
+        // test kafka config
+        String kafkaConfig = "akka.discovery.config.services.kafka.from-config=false";
+        String kafkaPort = "akka.discovery.config.services.kafka.endpoints[0].port" + KAFKA_CONTAINER.getFirstMappedPort();
```
dev930018:
```suggestion
        String kafkaPort = "akka.discovery.config.services.kafka.endpoints[0].port" + KAFKA_CONTAINER.getFirstMappedPort();
```
```suggestion
        String kafkaPort = "akka.discovery.config.services.kafka.endpoints[0].port=" + KAFKA_CONTAINER.getFirstMappedPort();
```

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
```diff
@@ -0,0 +1,196 @@
...
+@EnableCaching
+@ContextConfiguration(initializers = {RedisInitializer.class})
+@Import(RedisAutoConfiguration.class)
+@ActiveProfiles({"redis"})
+class UserServiceAsyncProxyErrorTest {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UserServiceAsyncProxyErrorTest.class);
+
+
+    private static final Function<Integer, String> REQUEST_FACTORY_JSON = i ->
```
dev930018: I get why we'd have this in contract tests, but why here?

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
```diff
@@ -0,0 +1,196 @@
...
+                .as("verify that exception is propagated into an auditable object and returned")
+                .isNotNull();
+
+        // Then check the error message contains the correct message
+        assertThat(auditableUserResponse.getAuditErrorMessage().getError().getMessage())
+                .as("verify that exception is propagated into an auditable object and returned")
+                .isEqualTo(NoSuchUserIdException.class.getName() + ": No userId matching test-user-id found in cache");
+    }
+
+    public static class RedisInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
```
dev930018: This shouldn't need redis. Use caffeine if you need a cache implementation for this test

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
```diff
@@ -230,7 +169,7 @@ void testRequestSet() {
                                 assertThat(result.headers().lastHeader(Token.HEADER).value())
                                         .isEqualTo(ContractTestData.REQUEST_TOKEN.getBytes())),
                 () -> assertThat(results.get(0).value().get("user").get("userId").get("id").asText())
-                        .isEqualTo(ContractTestData.USER_ID.getId()),
+                        .isEqualTo("test-user-id"),
```
dev930018: Suggest reverting to use ContractTestData

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/rest/HealthActuatorContractTest.java
```diff
@@ -33,8 +34,12 @@
  * This is done by checking the service is still alive and healthy by REST GET /actuator/health.
  * This should return 200 OK if the service is healthy.
  */
-@SpringBootTest(classes = UserApplication.class, webEnvironment = WebEnvironment.RANDOM_PORT)
-@ActiveProfiles({"caffeine", "akkatest"})
+@SpringBootTest(
+        classes = UserApplication.class,
+        webEnvironment = WebEnvironment.RANDOM_PORT,
+        properties = {"management.health.kafka.enabled=false", "spring.autoconfigure.exclude=org.springframework.boot.autoconfigure.data.redis.RedisAutoConfiguration"}
```
dev930018: This Redis exclusion shouldn't be required if using the caffeine profile

user-service/src/main/java/uk/gov/gchq/palisade/service/user/model/AuditMessage.java
```diff
@@ -49,28 +49,21 @@
...
-    protected final String serverIP;  //the server IP address for the service
-
+    protected final String serverIP;  // The server IP address for the service
     @JsonProperty("serverHostname")
-    protected final String serverHostname;  //the hostname of the server hosting the service
-
+    protected final String serverHostname;  // The hostname of the server hosting the service
     @JsonProperty("attributes")
-    protected final Map<String, Object> attributes;  //Map<String, Object> holding optional extra information
+    protected final Map<String, Object> attributes;  // Map<String, Object> holding optional extra information
```
dev930018, line 64: At some point, it's probably worth changing this, but you decide if wanted now
```suggestion
    protected final Map<String, String> attributes;  // Map<String, String> holding optional extra information
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Kafka health indicator. Check that the consumer group can be accessed and is registered with the cluster,
+ * if not mark the service as unhealthy.
+ */
+@Component("kafka")
+@ConditionalOnEnabledHealthIndicator("kafka")
+public class KafkaHealthIndicator implements HealthIndicator {
```
dev930018: Prefer explicit `@Bean` in `@Configuration` over `@Component`

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaHealthIndicator.class);
+    private final AdminClient adminClient;
+    @Value("${akka.kafka.consumer.kafka-clients.group.id}")
+    private String groupId;
```
dev930018: Prefer taking the config object (see somewhere in the `stream` package) as an argument to the constructor over using the `@Value` annotation

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+                .filter(healthy -> healthy)
+                .map(up -> Health.up().build())
+                .orElseGet(() -> Health.down().build());
+    }
+
+    private boolean performCheck() {
+        try {
+            Map<String, ConsumerGroupDescription> groupDescriptionMap = this.adminClient.describeConsumerGroups(Collections.singletonList(this.groupId))
+                    .all()
+                    .get(1, TimeUnit.SECONDS);
```
dev930018, line 86:
```suggestion
                    .join();
```
avoids the need for the catch

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+
+            if (consumerGroupDescription.state() == ConsumerGroupState.STABLE) {
+                boolean assignedGroupPartition = consumerGroupDescription.members().stream()
+                        .noneMatch(member -> (member.assignment() == null || member.assignment().topicPartitions().isEmpty()));
+                if (!assignedGroupPartition) {
+                    LOGGER.error("Failed to find kafka topic-partition assignments");
+                }
+                return assignedGroupPartition;
+            }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
```
dev930018, line 101: Suggest us COmpletableFuture::join and avoid the need for this catch

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+        this.adminClient = adminClient;
+    }
+
+    @Override
+    public Health getHealth(final boolean includeDetails) {
+        return Optional.of(performCheck())
+                .filter(healthy -> healthy)
+                .map(up -> Health.up().withDetail("group", this.groupId).build())
+                .orElseGet(() -> Health.down().withDetail("group", this.groupId).build());
+    }
```
dev930018: There's a lot more detail we *could* be adding here, but i think groupId is sufficient

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,106 @@
...
+                    LOGGER.error("Failed to find kafka topic-partition assignments");
+                }
+                return assignedGroupPartition;
+            }
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            LOGGER.warn("Timeout during Kafka health check for group {}", this.groupId, e);
+            Thread.currentThread().interrupt();
+            return false;
+        }
+        return true;
```
dev930018: By default it's healthy? Hmm...
```suggestion
        return false;
```

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
dev930018: This looks a bit weird, and the Optional is never empty so why use it? Why not:
```
if (performCheck()) {
  return Health.up().with...
} else {
  return Health.down().with...
}

File: user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaProducerService.java [hunk 35]
```diff
@@ -92,15 +90,21 @@ public KafkaProducerService(final Sink<ProducerRecord<String, UserRequest>, Comp
...
+                // Run the graph
                 .run(this.materializer)
-                .toCompletableFuture()
-                .join();
 
-        // Return results
-        return new ResponseEntity<>(HttpStatus.ACCEPTED);
+                // Return a CompletableFuture<Void> result
+                .toCompletableFuture()
+                .thenApply(x -> null);
```
dev930018, line 108: I assume returning a reasonable status code etc. is still done in the controller?

dev930018 (changes requested):
43/63 files viewed, will complete tomorrow

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/UserServiceCachingProxy.java
```diff
@@ -28,25 +28,40 @@
...
+public class UserServiceCachingProxy {
     private static final Logger LOGGER = LoggerFactory.getLogger(UserServiceCachingProxy.class);
     private final UserService service;
 
     /**
-     * Constructor for the {@link UserServiceCachingProxy}
+     * Default constructor used to create the UserServiceCachingProxy
      *
-     * @param service an implementation of the {@link UserService}
+     * @param service {@link UserService} this service calls
```
dev930018: I prefer the original documentation

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/UserServiceCachingProxy.java
```diff
@@ -28,25 +28,40 @@
...
      */
     public UserServiceCachingProxy(final UserService service) {
         this.service = service;
     }
 
+    /**
+     * Using the userId as the key, retrieves the {@link User} from the cache, if the User doesn't exist,
+     * then a message is Logged, and a {@link NullUserService} getUser() is called,
+     * which will throw a {@link uk.gov.gchq.palisade.service.user.exception.NoSuchUserIdException}.
```
dev930018: While this is true at the moment, it isn't in general: the service is called, which may or may not throw a NoSuchUserIdException

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/PropertiesConfigurer.java
```diff
@@ -54,12 +54,12 @@
  * Parse and convert Spring maps and lists to Akka configs
  */
 public class PropertiesConfigurer extends PropertySourcesPlaceholderConfigurer implements InitializingBean {
-    private static final Pattern INDEXED_PROPERTY_PATTERN = Pattern.compile("^\\s*(?<path>\\w+(?:\\.\\w+)*)\\[(?<index>\\d+)\\]\\.*(.*?)$");
+    private static final Pattern INDEXED_PROPERTY_PATTERN = Pattern.compile("^\\s*(?<path>\\w+(?:\\.\\w+)*)\\[(?<index>\\d+)]\\.*(.*?)$");
```
dev930018: Are you sure? We're trying to match a literal `[index]` so this `]` should be escaped as it was

user-service/src/main/resources/application.yaml
```diff
@@ -50,17 +42,15 @@ management:
...
-  client:
-    config:
-      default:
-        connectTimeout: 5000
-        readTimeout: 5000
-        loggerLevel: basic
-    default-config: default
-
-user-service:
-  implementation: null-user-service
```
dev930018, line 63: Do we no longer need this selector?

dev930018 (changes requested):

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/README.md
pd104923: Done

user-service/src/main/java/uk/gov/gchq/palisade/service/user/model/AuditMessage.java
pd104923: I think this should be left for now so that it's done on all the services at the same time.

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/rest/HealthActuatorContractTest.java
pd104923: Removed the exclusions

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
pd104923: Reverted

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
pd104923: Reworked this class to use Caffeine

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
pd104923: This isn't actually used properly as we return the same json String each time. Adjusted to just be a `String` value.

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaInitializer.java
pd104923: Done

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/RedisUserCachingTest.java
pd104923: Done

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: The `.join()` method is not available to use here as it's a `KafkaFuture`

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: These are necessary as we're dealing with `KafkaFuture`s, not `CompletableFuture`s.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: I think groupId is useful for now but you're right, we could add a lot more information here if required.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: The logic for this has been changed slightly so false is returned unless the try is successful.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: Changed

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: The topic configuration has been added to the constructor for this class. The "groupId" field is set within the `performCheck` method.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaProducerService.java
pd104923: Correct, the controller returns a `202` status code.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/UserServiceCachingProxy.java
pd104923: Reverted

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/UserServiceCachingProxy.java
pd104923: Generalised the JavaDoc

user-service/src/main/java/uk/gov/gchq/palisade/service/user/stream/PropertiesConfigurer.java
pd104923: Reverted

user-service/src/main/resources/application.yaml
pd104923: I think we still need this so have added it back in.

user-service/src/main/java/uk/gov/gchq/palisade/service/user/service/KafkaHealthIndicator.java
pd104923: A bean has been added to the ApplicationConfiguration.

user-service/README.md
```diff
@@ -1,26 +1,111 @@
...
+* `POST api/user`
+  - accepts an `x-request-token` `String` header, any number of extra headers, and a single `UserRequest` in the body
+  - returns a `202 ACCEPTED` after writing the headers and `AuditableUserResponse` to kafka
+* `POST api/user/multi`
+  - accepts an `x-request-token` `String` header, any number of extra headers, and a list of `UserRequest`s within the body
+  - returns a `202 ACCEPTED` after writing the headers and `AuditableUserResponse` to kafka
```
dev930018: It writes a UserRequest to kafka and returns, no?

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
```diff
@@ -0,0 +1,149 @@
...
+        classes = {ApplicationConfiguration.class, CacheAutoConfiguration.class},
+        webEnvironment = WebEnvironment.NONE,
+        properties = {"spring.cache.caffeine.spec=expireAfterWrite=1s, maximumSize=100"}
+)
+@EnableCaching
+@ActiveProfiles({"caffeine"})
+class UserServiceAsyncProxyErrorTest {
+
+    private static final String REQUEST_JSON = "{\"userId\":\"test-user-id\",\"resourceId\":\"/test/resourceId\",\"context\":{\"class\":\"uk.gov.gchq.palisade.Context\",\"contents\":{\"purpose\":\"purpose\"}}}";
+    private static final String NO_USER_REQUEST_JSON = "{\"userId\":\"Not-a-real-user\",\"resourceId\":\"/test/resourceId\",\"context\":{\"class\":\"uk.gov.gchq.palisade.Context\",\"contents\":{\"purpose\":\"purpose\"}}}";
```
dev930018: Why have this in a component test? Just create the objects, we don't care about the json

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
```diff
@@ -0,0 +1,149 @@
...
+        // Then check that there is an error message
+        assertThat(auditableUserResponse.getAuditErrorMessage())
+                .as("verify that exception is propagated into an auditable object and returned")
+                .isNotNull();
+
+        // Then check the error message contains the correct message
+        assertThat(auditableUserResponse.getAuditErrorMessage().getError().getMessage())
+                .as("verify that exception is propagated into an auditable object and returned")
+                .isEqualTo(NoSuchUserIdException.class.getName() + ": No userId matching Not-a-real-user found in cache");
+    }
```
dev930018: I'd prefer to check the `assertThat(error).hasType(NoSuchUserIdException.class)` rather than inspecting the `getMessage` string

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
```diff
@@ -271,36 +213,56 @@ void testNoSuchUserIdExceptionIsThrown() {
...
+                        .hasSize(1),
 
-                () -> assertThat(results.getLast().headers().lastHeader(StreamMarker.HEADER).value())
-                        .isEqualTo(StreamMarker.END.toString().getBytes())
-        );
+                // The error has the relevant headers, including the token
+                () -> assertThat(errorResults)
+                        .allSatisfy(result ->
+                                assertThat(result.headers().lastHeader(Token.HEADER).value())
+                                        .isEqualTo(ContractTestData.REQUEST_TOKEN.getBytes())),
```
dev930018: Combine these two related assertions:
```
assertThat(errorResults)
  .hasSize(1)
  .allSatisfy(result -> assertThat(headers are correct))
  .allSatisfy(result -> assertThat(value.get("error").get("message")).isEqualTo(...))
```

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
```diff
@@ -271,36 +213,56 @@ void testNoSuchUserIdExceptionIsThrown() {
...
 
-        // All but the first and last have the expected message
-        results.removeFirst();
-        results.removeLast();
-        assertThat(results).isEmpty();
+                // The error has a message that contains the throwable exception, and the message
+                () -> assertThat(errorResults.get(0).value().get("error").get("message").asText())
+                        .isEqualTo(NoSuchUserIdException.class.getName() + ": No userId matching invalid-user-id found in cache")
+        );
```
dev930018: Ditto, combine this with the above

dev930018 (changes requested):
Also need to clean up the pom a bit:
* remove embedded-redis in favour of testcontainers 
* remove spring-cloud-config-client and spring-boot-configuration-processor (i think they were used for feign/eureka stuff?)
* remove spring-boot-starter-jpa and h2
*

user-service/README.md
pd104923: You're quite right. Updated so that a UserRequest is sent to Kafka :D

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
pd104923: The issue with this is that the "Throwable" error is actually a `CompletionException` with a message that contains the NoSuchUserIdException class.
Hence why we have the classname and the message in the `isEqualTo` check.

user-service/src/component-tests/java/uk/gov/gchq/palisade/component/user/service/UserServiceAsyncProxyErrorTest.java
pd104923: Removed all the JSON stuff. Just creates the objects.

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
pd104923: Updated

user-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/user/kafka/KafkaContractTest.java
pd104923: Done

dev930018 (approved):

ac74475:
possible pipeline race condition being looked into, hold merge until it has been checked out

PR #217, opened 2020-12-03
Pal 864 palisade service auditing

pd104923:


palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
```diff
@@ -0,0 +1,133 @@
...
+package uk.gov.gchq.palisade.service.palisade.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * This class is a container for {@code PalisadeResponse} and {@code AuditErrorMessage} during stream processing.
```
m78233: Typo, should be @code PalisadeResponse

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
```diff
@@ -0,0 +1,133 @@
...
+        return Objects.equals(palisadeRequest, that.palisadeRequest) &&
+                Objects.equals(auditErrorMessage, that.auditErrorMessage);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(palisadeRequest, auditErrorMessage);
+    }
+
+    public PalisadeRequest getPalisadeRequest() {
```
m78233, line 78: Getters should be annotated with @Override

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * This class is a container for {@code PalisadeResponse} and {@code AuditErrorMessage} during stream processing.
+ * Under normal conditions only one of these will be non-null, indicating failed or successful processing.
+ */
+
+public final class AuditablePalisadeRequest {
```
m78233, line 32: Ought to have a corresponding unit test

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            /**
+             * Compose value
+             *
+             * @param request          value or null
+             * @param auditErrorMessage value or null
+             * @return value object
+             */
+            AuditablePalisadeRequest withResponseAndError(PalisadeRequest request, AuditErrorMessage auditErrorMessage);
```
m78233, line 129: Looked at your RunnableGraph, so not likely an issue, but worth asking, are you ever going to get an instance with both?

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Kafka health indicator. Check that the producer group can be accessed and is registered with the cluster,
+ * if not mark the service as unhealthy.
+ */
+@Component("kafka")
+@ConditionalOnEnabledHealthIndicator("kafka")
+public class KafkaHealthIndicator implements HealthIndicator {
```
m78233, line 44: How do we test this?

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
m78233, line 133: method is only meant to be used by other methods in this class so it should be private

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: This class does not extend or implement anything so adding the `@Override` annotation is redundant.
Did you mean to say "use the `@Generated` annotation"?

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/web/PalisadeRestController.java
```diff
@@ -81,7 +77,7 @@ public PalisadeRestController(final PalisadeService palisadeService, final Error
             LOGGER.error("PalisadeRestController Exception thrown", e);
             httpStatus = HttpStatus.INTERNAL_SERVER_ERROR;
             Map<String, Object> attributes = new HashMap<>();
-            errorHandlingService.createErrorMessage(request, attributes, e);
+            palisadeService.errorMessage(request, token, attributes, e);
         }
         return ResponseEntity.status(httpStatus)
```
m78233, line 82: palisadeResponse can be null when registerDataRequest is called and so token can also be null. These need to be handled in sending the errorMessage to the Audit plus the response back to the client.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: I'm not sure what the typo is here. Are you referring to the braces `{}`?

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/PalisadeService.java
```diff
@@ -58,27 +61,52 @@ public PalisadeService(final Materializer materializer) {
...
+     */
+    public void errorMessage(final PalisadeRequest request, final String token,
+                             final Map<String, Object> attributes, final Throwable error) {
+        // Sends the information to the "error" topic
+        // We need to include the PalisadeRequest information and the Error that occurred.
+        AuditErrorMessage errorMessage = AuditErrorMessage.Builder.create(request, attributes).withError(error);
+        AuditablePalisadeRequest auditableRequest = AuditablePalisadeRequest.Builder.create().withAuditErrorMessage(errorMessage);
+        TokenRequestPair requestPair = new TokenRequestPair(token, auditableRequest);
+
+        futureSink.thenApply(sink -> Source.single(requestPair).runWith(sink, materializer));
```
m78233, line 107: I think thenAccept is more appropriate here.

m78233 (changes requested):
Like your style :) .  There are few comments, double check on a few thing and bit of code that needs to be reworked.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: Created a component test class with the other request test classes.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/KafkaHealthIndicator.java
pd104923: I'm not sure that this can be tested properly within the code. I'll mark as resolved for now.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: Private methods within Interfaces need to have a body but this is not necessary here. Leaving as is.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/web/PalisadeRestController.java
pd104923: So the error is passed to the `error` topic and I don't think we need to passing the details of the internal errors to the client as the token has not been created.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/PalisadeService.java
pd104923: Using a `thenAccept` might be better for a void method as we don't need to use the returned value from a `thenApply` call. Leaving for now but I am happy to use either.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: Realised the silly mistake, updated to `@PalisadeRequest`

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/model/AuditablePalisadeRequest.java
pd104923: Added `@Generated` annotations.

m78233 (dismissed):
Good, covers off what we discussed.

palisade-service/src/component-tests/java/uk/gov/gchq/palisade/component/palisade/model/AuditErrorMessageTest.java
```diff
@@ -69,27 +69,27 @@ void testAuditErrorMessageSerialisingAndDeserialising() throws IOException {
         ObjectContent<AuditErrorMessage> auditErrorMessageObjectContent = jsonTester.parse(auditErrorMessageJsonContent.getJson());
         AuditErrorMessage auditErrorMessageObject = auditErrorMessageObjectContent.getObject();
 
-        assertAll("AuditSerialisingDeseralisingAndComparison",
-                () -> assertAll("AuditSerialisingComparedToString",
+        assertAll("AuditErrorMessage Serialising and Deseralising Comparison",
+                () -> assertAll("AuditErrorMessage Serialising Compared To String",
```
dbz120896, line 73: Much better to create an "expected" instance and then use:
```
assertThat(actual)
            .usingRecursiveComparison()
            .isEqualTo(expected);
```
This means that if a new field is added, the test will automatically pick it up and include it in the comparison. This is good practice for objects that are used for JSON (de)serialisation.

It also makes the tests much easier to read.

palisade-service/src/component-tests/java/uk/gov/gchq/palisade/component/palisade/model/PalisadeResponseTest.java
```diff
@@ -46,22 +46,22 @@
...
                         () -> assertThat(responseJsonContent).extractingJsonPathStringValue("$.token").isEqualTo("tokenID")
                 ),
-                () -> assertAll("AuditDeserialisingComparedToObject",
+                () -> assertAll("Palisade Response Deserialising Compared To Object",
                         () -> assertThat(palisadeResponseObject.getToken()).isEqualTo(palisadeResponse.getToken())
                 ),
-                () -> assertAll("ObjectComparison",
+                () -> assertAll("Object Comparison",
                         //compares the two objects using the objects equal method
                         () -> assertThat(palisadeResponseObject).usingRecursiveComparison().isEqualTo(palisadeResponse)
```
dbz120896: This does not use object equals. This will match fields by name and test then test for equality. This is then performed recursively. This is the only test that you need to do, apart from adding `assertThat(palisadeResponseObject).isEqualTo(palisadeResponse)`. This will then test the `equals` implementation. All the other individual field test can then be removed.

dbz120896 (changes requested):
Just a couple of changes to improve tests

palisade-service/src/component-tests/java/uk/gov/gchq/palisade/component/palisade/model/AuditErrorMessageTest.java
pd104923: This type of test isn't going to work for checking the string values as you can't do a recursive comparison and there will be differences in the throwable error fields. The object comparison at the end should help to flag any new fields that have been added.

palisade-service/src/component-tests/java/uk/gov/gchq/palisade/component/palisade/model/PalisadeResponseTest.java
pd104923: I've removed the recursive comparison so that it uses the `equals` method.

dbz120896 (dismissed):

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/KafkaHealthIndicator.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(KafkaHealthIndicator.class);
+    private final AdminClient adminClient;
+    private static final List<String> TOPIC_NAMES = Arrays.asList("request", "error");
```
dev-958: there really is no need for these to be hardcoded as they can be extracted from the configuration, it would leave the code less fragile.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/stream/config/AkkaRunnableGraph.java
```diff
@@ -60,16 +63,37 @@
...
 
         return source
                 .flatMapConcat((TokenRequestPair tokenAndRequest) -> {
+
+                    //decidePartition
+                    Integer partition = Token.toPartition(tokenAndRequest.first(), outputTopic.getPartitions());
+
+                    BiFunction<AuditablePalisadeRequest, Headers, ProducerRecord<String, byte[]>> recordFunc = (AuditablePalisadeRequest value, Headers headers) -> {
+                        // Make the AuditablePalisadeRequest an Optional
+                        Optional<AuditablePalisadeRequest> auditablePalisadeRequest = Optional.ofNullable(value);
```
dev-958, line 84: I don't understand why you would need an Auditable... here as there are no service calls in the flow that could throw an exception. The only place exceptions could occur is when marshaling the input message into a java object but you (quite rightly) handle that in the controller and write to the audit topic there.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/stream/config/AkkaRunnableGraph.java
```diff
@@ -60,16 +63,37 @@
...
+                        // Map the auditable request to either a PalisadeRequest or AuditErrorMessage
+                        return auditablePalisadeRequest.map(AuditablePalisadeRequest::getAuditErrorMessage).map(audit ->
+                                new ProducerRecord<>(errorTopic.getName(), partition, (String) null,
+                                                SerDesConfig.errorValueSerializer().serialize(null, audit), headers))
```
dev-958, line 88: I don't see how this could ever happen as you handle exceptions in the controller and none can happen here.

dev-958 (changes requested):
See comments inline.

Also, could we have a web test that verifies that the JSON response payload returned when an error occurs is not gobbledygook as it was pre-streaming?

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/service/KafkaHealthIndicator.java
pd104923: The values for the topic list are no longer hard-coded

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/stream/config/AkkaRunnableGraph.java
pd104923: The error doesn't get written to the error topic from the controller. It's passed to the `PalisadeService` class which then creates the Auditable..... object. This is then added to the sink which is then processed by the `RunnableGraph`.

palisade-service/src/main/java/uk/gov/gchq/palisade/service/palisade/stream/config/AkkaRunnableGraph.java
pd104923: This is where the encountered error is added to the `error` topic. If it can be done a different way then I'm happy to discuss it.

dev-958 (approved):

PR #218, opened 2020-12-08
Pal 866 policy service error handling

m78233:


policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaRunnableGraph.java
```diff
@@ -70,68 +73,64 @@ KafkaProducerService kafkaProducerService(
...
-                                .thenApply(rules -> PolicyServiceHierarchyProxy.applyRulesToResource(policyRequest.getUser(), policyRequest.getResource(), policyRequest.getContext(), rules))
-                                // If this is a proper request (not start or end of stream)
-                                // If this is empty it could be a start/end of stream, or it could be a redacted resource, so keep track of both cases
-                                .thenApply(resource -> new Pair<>(messageAndRequest.first(), Optional.of(new Pair<>(policyRequest, Optional.ofNullable(resource))))))
-                        // If is a START/END of stream then treat as accessible
-                        .orElse(CompletableFuture.completedFuture(new Pair<>(messageAndRequest.first(), Optional.empty()))))
+                .mapAsync(PARALLELISM, messageAndRequest -> service.getResourceRules(messageAndRequest.second())
+                        //need to change this only apply if we have rules and if we have a PolicyRequest
+                        .thenApply(auditable -> PolicyServiceHierarchyProxy.applyRulesToResource(auditable))
```
pd104923: Can be replaced with a method reference
`.thenApply(PolicyServiceHierarchyProxy::applyRulesToResource)`

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaRunnableGraph.java
```diff
@@ -70,68 +73,64 @@ KafkaProducerService kafkaProducerService(
...
-                        // If this is a start or end of stream
-                        .orElse(CompletableFuture.completedFuture(Optional.empty()))
-                        // Keep track of original committable
-                        .thenApply(response -> new Pair<>(messageRequestResource.first(), response)))
+                // Get the record level rules for all resources that weren't course grain filtered
+                .mapAsync(PARALLELISM, (Pair<CommittableMessage<String, PolicyRequest>, AuditablePolicyResourceResponse> messageAndModifiedRequest) ->
+                        service.getRecordRules(messageAndModifiedRequest.second())
+                                //check to see if the first service request threw an exception and if so deal with it
+                        .thenApply(modifiedResource -> modifiedResource.chain(messageAndModifiedRequest.second().getAuditErrorMessage()))
+                                .thenApply(response -> Pair.create(messageAndModifiedRequest.first(), response))
```
pd104923: The definition of the lambda element isn't strictly needed as there are no curly braces `{}` around the lambda code, unless sq says otherwise.
```suggestion
                .mapAsync(PARALLELISM, messageAndModifiedRequest ->
                        service.getRecordRules(messageAndModifiedRequest.second())
                                //check to see if the first service request threw an exception and if so deal with it
                                .thenApply(modifiedResource -> modifiedResource.chain(messageAndModifiedRequest.second().getAuditErrorMessage()))
                                .thenApply(response -> Pair.create(messageAndModifiedRequest.first(), response))
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaRunnableGraph.java
```diff
@@ -70,68 +73,64 @@ KafkaProducerService kafkaProducerService(
...
+                            // Found an application error, produce an error message to be sent to the Audit service
+                            .map(audit -> ProducerMessage.single(
+                                    new ProducerRecord<>(errorTopic.getName(), requestRecord.partition(), requestRecord.key(),
+                                            SerDesConfig.errorValueSerializer().serialize(null, audit), requestRecord.headers()),
+                                    (Committable) messageAndResponse.first().committableOffset()))
+                            //Found a response message, produce a policy message to be sent to the output
+                            .orElse(ProducerMessage.single(
+                                    new ProducerRecord<>(outputTopic.getName(), requestRecord.partition(), requestRecord.key(),
+                                            SerDesConfig.ruleValueSerializer().serialize(null, messageAndResponse.second().getPolicyResponse()), requestRecord.headers()),
+                                    commitable));
```
pd104923: There's no need to cast to Committable for the variable. Plus, if we already have this then why are we not using it when creating the error ProcuderMessage?
```suggestion
                .map((Pair<CommittableMessage<String, PolicyRequest>, AuditablePolicyRecordResponse> messageAndResponse) -> {
                    ConsumerRecord<String, PolicyRequest> requestRecord = messageAndResponse.first().record();
                    Committable committable = messageAndResponse.first().committableOffset();
                    return Optional.ofNullable(messageAndResponse.second().getAuditErrorMessage())
                            // If there is an application error, produce an error message to be sent to the error topic
                            .map(audit -> ProducerMessage.single(
                                    new ProducerRecord<>(errorTopic.getName(), requestRecord.partition(), requestRecord.key(),
                                            SerDesConfig.errorValueSerializer().serialize(null, audit), requestRecord.headers()),
                                    committable))
                            // If there is a response message, produce a policy response to be sent to the output topic
                            .orElse(ProducerMessage.single(
                                    new ProducerRecord<>(outputTopic.getName(), requestRecord.partition(), requestRecord.key(),
                                            SerDesConfig.ruleValueSerializer().serialize(null, messageAndResponse.second().getPolicyResponse()), requestRecord.headers()),
                                    committable));
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
```diff
@@ -51,7 +50,9 @@
 @Configuration
 public class AkkaComponentsConfig {
     private static final StreamComponents<String, PolicyRequest> INPUT_COMPONENTS = new StreamComponents<>();
-    private static final StreamComponents<String, PolicyResponse> OUTPUT_COMPONENTS = new StreamComponents<>();
+    private static final StreamComponents<String, byte[]> OUTPUT_COMPONENTS = new StreamComponents<>();
+
+    //no in the attribute masking service why do we have it
     private static final StreamComponents<String, AuditErrorMessage> ERROR_COMPONENTS = new StreamComponents<>();
```
pd104923: This is no longer required and should be removed.
The `plainErrorSink` is also not needed anymore and can be removed.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
```diff
@@ -80,11 +81,12 @@
...
     @Bean
-    Sink<Envelope<String, PolicyResponse, Committable>, CompletionStage<Done>> committableResponseSink(final ActorSystem actorSystem) {
-        ProducerSettings<String, PolicyResponse> producerSettings = OUTPUT_COMPONENTS.producerSettings(
+    Sink<Envelope<String, byte[], Committable>, CompletionStage<Done>> committableResponseSink(final ActorSystem actorSystem) {
+        ProducerSettings<String, byte[]> producerSettings = OUTPUT_COMPONENTS.producerSettings(
                 actorSystem,
                 SerDesConfig.ruleKeySerializer(),
-                SerDesConfig.ruleValueSerializer());
+               // SerDesConfig.ruleValueSerializer());
+                SerDesConfig.passthroughValueSerializer());
```
pd104923: Can this commented line be deleted please, it's not needed.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
```diff
@@ -51,7 +50,9 @@
 @Configuration
 public class AkkaComponentsConfig {
```
pd104923, line 57: A bean needs to be created in this class for the Kafka AdminClient. This can be seen in the AMS code.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/SerDesConfig.java
```diff
@@ -120,6 +120,15 @@ private SerDesConfig() {
         };
     }
 
+    /**
+     * Kafka value serialiser for downstream messages going out as output
+     *
+     * @return an appropriate value serialiser for the topic's message content (AttributeMaskingResponse)
```
pd104923:
```suggestion
     * @return an appropriate value serialiser for the topic's message content (PolicyResponse)

File: policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceHierarchyProxy.java [hunk 8]
```diff
@@ -81,6 +84,28 @@ public PolicyServiceHierarchyProxy(final PolicyServiceCachingProxy service) {
         return mergedRules;
     }
 
+    /**
+     *  Applies the {@link Rules} to the {@link Resource}.
+     * @param auditablePolicyResourceResponse container for holding the {@code PolicyRequest}, {@code Rules} and the
+     * {@code AuditErrorMessage}
+     * @return either the same instance of {@code AuditablePolicyResourceResponse} or a new instance with the
+     * {@code Resource modified}
+     */
```
pd104923:
```suggestion
    /**
     * Applies the {@link Rules} to the {@link Resource}.
     *
     * @param auditablePolicyResourceResponse container for holding the {@code PolicyRequest}, {@code Rules} and the {@code AuditErrorMessage}
     * @return either the same instance of {@code AuditablePolicyResourceResponse} or a new instance with the modified {@code Resource}
     */
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxy.java
```diff
@@ -48,22 +55,63 @@ public PolicyServiceAsyncProxy(final PolicyServiceHierarchyProxy service, final
...
+     * @param modifiedAuditable the resource the user wants access to
      * @return the record rules for the LeafResource, if null then an exception will be thrown
      */
-    public CompletableFuture<Rules<Serializable>> getRecordRules(final LeafResource resource) {
-        return CompletableFuture.supplyAsync(() -> service.getRecordRules(resource), executor);
+    public CompletableFuture<AuditablePolicyRecordResponse> getRecordRules(final AuditablePolicyResourceResponse modifiedAuditable) {
+
+       PolicyRequest nullableRequest = modifiedAuditable.getPolicyRequest();
+        return Optional.ofNullable(nullableRequest)
+                .map((PolicyRequest request) ->
```
pd104923:
```suggestion
                .map(request ->
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxy.java
```diff
@@ -48,22 +55,63 @@ public PolicyServiceAsyncProxy(final PolicyServiceHierarchyProxy service, final
...
-     * @param resource the resource the user wants access to
+     * @param nullableRequest the resource the user wants access to
      * @return the rules for the LeafResource, if null then an exception will be thrown
      */
-    public CompletableFuture<Rules<LeafResource>> getResourceRules(final LeafResource resource) {
-        return CompletableFuture.supplyAsync(() -> service.getResourceRules(resource), executor);
+    public CompletableFuture<AuditablePolicyResourceResponse> getResourceRules(final @Nullable PolicyRequest nullableRequest) {
+
+        return Optional.ofNullable(nullableRequest)
+                .map((PolicyRequest request) ->
```
pd104923:
```suggestion
                .map(request ->
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxy.java
```diff
@@ -48,22 +55,63 @@ public PolicyServiceAsyncProxy(final PolicyServiceHierarchyProxy service, final
...
+                        CompletableFuture.supplyAsync(() -> service.getRecordRules(request.getResource()))
+                                .thenApply(rules -> AuditablePolicyRecordResponse.Builder.create()
+                                        .withPolicyResponse(PolicyResponse.Builder.create(request)
+                                                .withRules(rules))
+                                        .withNoErrors())
+                                .exceptionally(e -> AuditablePolicyRecordResponse.Builder.create()
+                                        .withPolicyResponse(PolicyResponse.Builder.create(request)
+                                                .withRules(new Rules()))
```
pd104923, line 106: Can this not have a `null` value?

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
```diff
@@ -84,20 +81,56 @@ public String getResourceId() {
     }
 
     @Generated
-    public Context getContext() throws JsonProcessingException {
-        return MAPPER.treeToValue(this.context, Context.class);
+    public Context getContext()  {
+        try {
+            return MAPPER.treeToValue(this.context, Context.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Failed to get Context", e);
```
pd104923: Would recommend throwing a Serialisation error rather than the general Runtime error. Maybe something like this but there might be a better one out there
```suggestion
            throw new SerializationFailedException("Failed to get Context", e);
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
```diff
@@ -84,20 +81,56 @@ public String getResourceId() {
...
     }
 
     @Generated
-    public User getUser() throws JsonProcessingException {
-        return MAPPER.treeToValue(this.user, User.class);
+    public User getUser()  {
+        try {
+            return MAPPER.treeToValue(this.user, User.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Failed to get User", e);
```
pd104923: Same as above
```suggestion
            throw new SerializationFailedException("Failed to get User", e);
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
```diff
@@ -84,20 +81,56 @@ public String getResourceId() {
...
+    JsonNode getUserNode() {
+        return this.user;
     }
 
+
+    public LeafResource getResource() {
+        try {
+            return MAPPER.treeToValue(this.resource, LeafResource.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException("Failed to get Resource", e);
```
pd104923: Same as above
```suggestion
            throw new SerializationFailedException("Failed to get Resource", e);
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
```diff
@@ -50,11 +51,7 @@
     private final JsonNode context;  // Json Node representation of the Context
     private final JsonNode user;  //Json Node representation of the User
     private final JsonNode resource; // Json Node representation of the Resources
-
-    /**
-     * Holds all of the Rules applicable to this request
-     */
-    public final Rules rules;
+    private final Rules rules;  //Holds all of the Rules applicable to this request
```
pd104923, line 55: Do we not want this field to be public anymore? I thought this was done so we could update/use this field within the policy-service?

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
```diff
@@ -78,11 +77,11 @@ public String getResourceId() {
     }
 
     @Generated
-    public Context getContext() {
+    public Context getContext()  {
         try {
             return MAPPER.treeToValue(this.context, Context.class);
         } catch (JsonProcessingException e) {
-            throw new SerializationFailedException("Failed to get Context", e);
+            throw new RuntimeException("Failed to get Context", e);
```
pd104923: Why is this being changed to a `RuntimeException`?

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
```diff
@@ -96,7 +95,7 @@ public User getUser() {
         try {
             return MAPPER.treeToValue(this.user, User.class);
         } catch (JsonProcessingException e) {
-            throw new SerializationFailedException("Failed to get User", e);
+            throw new RuntimeException("Failed to get User", e);
```
pd104923: Same here

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
```diff
@@ -110,7 +109,7 @@ public LeafResource getResource() {
         try {
             return MAPPER.treeToValue(this.resource, LeafResource.class);
         } catch (JsonProcessingException e) {
-            throw new SerializationFailedException("Failed to get Resource", e);
+            throw new RuntimeException("Failed to get Resource", e);
```
pd104923: Same here

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/AuditablePolicyResourceResponse.java
```diff
@@ -0,0 +1,167 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import uk.gov.gchq.palisade.Generated;
+import uk.gov.gchq.palisade.resource.Resource;
+import uk.gov.gchq.palisade.rule.Rules;
+
+/**
+ * This class is a container for {@link PolicyRequest}, {@link Rules} and {@link AuditErrorMessage} during stream
+ * processing.  Under normal conditions {@code PolicyRequest} and { @code Rules} will be non-null, indicating
+ * successful process or { @code AuditErrorMessage} when there has been an error in the process.  The class can be in
```
pd104923:
```suggestion
 * processing.  Under normal conditions {@code PolicyRequest} and {@code Rules} will be non-null, indicating
 * successful process or {@code AuditErrorMessage} when there has been an error in the process.  The class can be in
```

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/AuditMessage.java
```diff
@@ -45,6 +45,7 @@
 @JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
 public class AuditMessage {
 
+
```
pd104923: Can this extra line be removed please.

policy-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/policy/kafka/KafkaContractTest.java
```diff
@@ -412,6 +437,8 @@ void testResourceRulesRedactedResource() {
         assertThat(results).isEmpty();
     }
 
+
+
```
pd104923: Remove the extra lines please

policy-service/src/unit-tests/java/uk/gov/gchq/palisade/service/policy/ApplicationTestData.java
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package uk.gov.gchq.palisade.service.policy;
```
pd104923, line 16: This class contains data for unit-test so there is no need to have the header setup or the record factory field as these are only needed for the kafka contract tests.
We only really need the objects used within the uni-tests here.

policy-service/src/unit-tests/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxyTest.java
```diff
@@ -0,0 +1,162 @@
...
+     * @throws Exception if the test fails
+     */
+    @Test
+    void tesGetRecordRulesWithNoPolicyRecord() throws Exception {
+        when(hierarchyProxy.getRecordRules(any())).thenThrow(new NoSuchPolicyException("Test"));
+
+        CompletableFuture<AuditablePolicyRecordResponse> completableFuture = asyncProxy.getRecordRules(AUDITABLE_POLICY_RESOURCE_RESPONSE_WITH_NO_RULES);
+        AuditablePolicyRecordResponse response = completableFuture.get();
+        assertThat(response.getPolicyResponse()).isNotNull();
+        assertThat(response.getAuditErrorMessage()).isNotNull();
```
pd104923, line 157: This doesn't seem right. The result of the test in the JavaDoc suggests that the policyResponse will be null, especially if an exception has been thrown by the `getRecordRules` method.

policy-service/pom.xml
```diff
@@ -155,6 +155,10 @@
             <version>${alpakka.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.aspectj</groupId>
+            <artifactId>aspectjweaver</artifactId>
+        </dependency>
```
pd104923: Is this dependency needed for anything?

pd104923 (changes requested):
A few things to change and there needs to be a KafkaHealthIndicator class so we can check the service can check Kafka.

policy-service/pom.xml
m78233: Not anymore.  I was originally using AspectJ for a Aspect Oriented Programming approach to handling the errors.  Changed the approach to latch it onto the proxy.  I'll removed thIS.

policy-service/src/contract-tests/java/uk/gov/gchq/palisade/contract/policy/kafka/KafkaContractTest.java
m78233: Done

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
m78233: This is part of the lingering issue with the getters.  When they were first built they declared throwing a Checked exception, JsonProcessingException.  In some point after that this one was modified by warping it in a Unchecked exception, SerializationFailedException.    Two  problems with this.   The first is that the SerializationFailedException is a Spring class and we are not meant to add any Spring classes to the message types.  Second the change is for this one class and we have the same type of method peppered in our code with different ways of operating.  That means that we will need different ways of handling the same code in different ways.  To address this, I raised a ticket, PAL-945.  The RuntimeException is a temporary measure as it is Unchecked and not a Spring class.  In my opinion, it should be a new Exception class fit for purpose, but that is the topic of the ticket.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
m78233: see previous message

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
m78233: see previous message

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
m78233: I though that is was a mistake. We should not be updating the content of an immutable object.  You want to change something, you create a new object.  If you disagree, I will run it up the flagpole.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
m78233: I have looked.  The SerializationFailedException is a Spring class so in this case it is a non-starter.  I think we need to create an Exception fit for purpose.  See the ticket PAL-946 Standard coding to handle Checked Exceptions

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
m78233: See comment above

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
m78233: See comment above

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxy.java
m78233: If you look at the service call in this block, service.getRecordRules(Resource). with throw a NoSuchPolicyException, (rules will be a null in this case) and the follow through will be the exceptionally instead of the thenApply.  In other words, the thenApply path will always have an instance of rules.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
m78233: Doubled check and realised that I don't create the AdminClient bean and the KafkaHealthIndicator class that goes with it.  I'll implement both.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
m78233: done

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaComponentsConfig.java
m78233: done

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaRunnableGraph.java
m78233: ok, looks easier to read

policy-service/src/unit-tests/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxyTest.java
m78233: Good point,  I bounced back and for on this one and the JavaDoc was not updated.  In the end, I found I needed  the PolicyResponse, but with an empty Rules for the code to work.  I'll update the JavaDoc to make it right.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
pd104923: Ok, happy for this to be changed.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyRequest.java
pd104923: Happy for this to be left as a `RuntimeException` for now as it will be addressed in a future ticket.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/model/PolicyResponse.java
pd104923: Happy for this to be left as a `RuntimeException` for now.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/service/PolicyServiceAsyncProxy.java
pd104923: Okay, that makes sense.

policy-service/src/main/java/uk/gov/gchq/palisade/service/policy/stream/config/AkkaRunnableGraph.java
developer01189998819991197253: resolved

PR #239, opened 2021-01-27
Pal 929 dockerfile lightweight images

dev930018:
Changed ~600MB openjdk image to ~200MB jre image (or ~400MB jdk on configurable maven property)


Jenkinsfile
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-@Library('jenkinsfile-lib')_
+@Library('jenkinsfile-lib@PAL-929-dockerfile-lightweight-images')_
```
kg981167224: Need to revert this before Merging, but after merging the Jenkinslib

kg981167224 (changes requested):

Jenkinsfile
dev930018: done

dev-958 (approved):

PR #249, opened 2021-02-10
PAL-965 set akka timeout to 300s

dev930018:
Ticket suggested `akka.kafka.consumer.close-timeout: infinite` but this doesn't seem to work.
Nominally set to a 300s timeout (up from 20s) which is long enough k8s would restart it, but this could be set longer.


kg981167224 (approved):

developer01189998819991197253 (approved):
Approved, although pipeline is still failing on deploy

dev-958 (approved):

PR #250, opened 2021-02-10
PAL-988 removed post-install from already-waiting create-topics job

dev930018:
Removed the post-install hook in favor of the default on-install mechanism. This should fix the pipeline deadlock problem.

This further offers the opportunity to use the `--wait` flag when setting up for the k8s smoke test to ensure all pods are ready before running


developer01189998819991197253 (dismissed):

charts/create-kafka-queues/templates/job.yaml
```diff
@@ -24,7 +24,6 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
   annotations:
```
dev-958: you don't need this annotation section at all then.

dev-958 (changes requested):

kg981167224 (approved):

developer01189998819991197253 (approved):

dev-958 (approved):

PR #258, opened 2021-02-24
PAL-1005 added wait-for-services script

dev930018:


Jenkinsfile
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-@Library('jenkinsfile-lib')_
+@Library('jenkinsfile-lib@PAL-1005-create-kafka-queues-delete-job')_
```
kg981167224, line 17: Remove before Merge.

create-kafka-queues/wait-for-services.sh
```diff
@@ -0,0 +1,30 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+wait_for_service() {
+  name=$(echo $1 | cut -f 1 -d :)
+  dest=$(echo $1 | cut -f 2 -d :)
+  port=$(echo $1 | cut -f 3 -d :)
+  until nc -vz $dest $port > /dev/null 2>&1; do
```
kg981167224, line 20: Would we want a timeout and appropriate error message for this. Since if it hangs it will tie up resources until it is manually killed.

kg981167224 (dismissed):
Just check if you agree with comments.

charts/create-kafka-queues/values.yaml
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-exports: {}
+exports:
   # zookeeper connection url
-  # zookeeper: palisade-zookeeper:2181
+  zookeeper: palisade-zookeeper:2181
```
dev-958: Seems a bit dodgy hard-coding the port in here, or even the zookeeper name. Can we not compose these in a _helpers template function using the values provided in the origin charts?

charts/create-kafka-queues/values.yaml
```diff
@@ -47,9 +47,17 @@ kafkaCreateTopics:
     replicationFactor: 1
     partitions: 1
 
+waitForServices:
```
dev-958, line 46: Again, I would compose these somehow. Too easy to break it.

create-kafka-queues/src/main/resources/values.yaml
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-exports: {}
+exports:
   # zookeeper connection url
-  # zookeeper: palisade-zookeeper:2181
+  zookeeper: palisade-zookeeper:2181
```
dev-958: same.

create-kafka-queues/src/main/resources/values.yaml
```diff
@@ -47,6 +47,14 @@ kafkaCreateTopics:
     replicationFactor: 1
     partitions: 1
 
+waitForServices:
```
dev-958, line 46: again

dev-958 (changes requested):
See comments. I think values from other charts need to be used so that changing ports or names don't break this.

dev-958 (approved):

PR #260, opened 2021-03-02
Pal 823 jvm performance tests

pd104923:


resource-service/src/main/resources/application-static.yaml
```diff
@@ -41,4 +41,4 @@ akka:
 # Connection detail for SimpleResourceService resources
 web:
   client:
-    data-service: http://localhost:8082
+    data-service: localhost:8082
```
m78233, line 44: Change aside, why does the resource-service need to have a URL for the data-service?

m78233 (commented):
Question about the need for a URL in resource-service for the data-service.

resource-service/src/main/resources/application-static.yaml
pd104923: This is needed for the `ConnectionDetail` that's added to the resource object.

m78233 (dismissed):

dbz120896 (dismissed):

dev-958 (approved):

PR #266, opened 2021-03-05
PAL-964, Updated IdFor in CompletenessEntity to be more unique

developer01189998819991197253:


resource-service/src/main/java/uk/gov/gchq/palisade/service/resource/domain/CompletenessEntity.java
```diff
@@ -101,14 +102,15 @@ public String getEntityId() {
...
      * @param entityType the Entity type enum object
-     * @param entityId   The Id of the entity, which eventually becomes a hash of the type and Id as a primary key
-     * @return the hash value used for the id, made up of the entity type and entity id values
+     * @param entityId   The Id of the entity, which eventually becomes a hashCode
+     * @return the hash value used for the id, made up of the length of EntityType enum,
+     * times the hashCode of entityId and the ordinal value of the passed in type
      */
     public static Integer idFor(final EntityType entityType, final String entityId) {
-        return Objects.hash(entityType, entityId);
+        return EntityType.values().length * entityId.hashCode() + entityType.ordinal();
```
kg981167224, line 113: Don't think this will work, entityId.hashcode returns an int, so Integer.MAX_VALUE * or + anything will overflow.
Returning a -ve id may work but not sure it's a good idea.

I thought we suggested that the fix was going to be String entityType+entityId.

kg981167224 (changes requested):

kg981167224 (approved):
After chatting, this is good, since the resource will be further refined by using the query ID

dev-958 (approved):

PR #270, opened 2021-03-10
Pal 760 update data service tests

m78233:


data-service/src/component-tests/java/uk/gov/gchq/palisade/component/data/common/CommonTestData.java
```diff
@@ -104,4 +105,29 @@ private CommonTestData() {
             .withToken(TOKEN)
             .withSuccessMessage(AUDIT_SUCCESS_MESSAGE)
             .withAuditErrorMessage(null);
+
```
kg981167224, line 108: Was this move needed? Since this data is only used in a single test.

kg981167224 (approved):
One minor comment

data-service/src/component-tests/java/uk/gov/gchq/palisade/component/data/common/CommonTestData.java
m78233: That was my take on these initially.  Now I am seeing that they are useful as a best practice in that it builds a library of existing test data and encourages the next person to goto the test class instead of re-creating the same the same somewhere else. Added bonus, reduces the clutter in the test class.

kg981167224 (dismissed):

dev930018 (dismissed):

developer01189998819991197253 (dismissed):

dev-958 (approved):

PR #274, opened 2021-03-15
PAL-489 Added ingress for FRS and DaemonSet for Traefik

dbz120896:


pd104923 (dismissed):

charts/filtered-resource-service/templates/ingress-route.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: traefik.containo.us/v1alpha1
+kind: IngressRoute
+metadata:
+  name: {{ include "filtered-resource-service.name" . }}
+  namespace: {{ include "palisade.namespace" . }}
+spec:
+  entryPoints:
+    - web
+  routes:
+    - match: Host(`{{ .Values.global.hostname }}`) && PathPrefix(`{{ include "filtered-resource-service.root" . }}/filteredResource`)
```
dev-958, line 24: using the service name in mixed case as the route url is a terrible idea and kind of exposes function, I recommend making this just /resource

charts/filtered-resource-service/templates/pathparam-regex.yaml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+#
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: filtered-resource-pathparam-regex
+  namespace: {{ include "palisade.namespace" . }}
+spec:
+  replacePathRegex:
+    regex: ^{{ include "filtered-resource-service.root" . }}/filteredResource/(.*)
```
dev-958, line 22: using the service name in mixed case as the route url is a terrible idea and kind of exposes function, I recommend making this just /resource

dev-958 (changes requested):

charts/filtered-resource-service/templates/ingress-route.yaml
dbz120896: No problem. I kept the ingress path the same as the routed path for now, but I can change them. It would look a bit odd though as we'll have `ws://172.17.0.3:32733/resource/resource/%25t`.

May be better to have 3 letter service names externally? So we'd get:

- http://172.17.0.3:32733/pal/......
- ws://172.17.0.3:32733/frs/......
- http://172.17.0.3:32733/dta/.....

This would really differentiate ingress URI's from internal.

charts/filtered-resource-service/templates/pathparam-regex.yaml
dbz120896: No problem. I kept the ingress path the same as the routed path for now, but I can change them. It would look a bit odd though as we'll have `ws://172.17.0.3:32733/resource/resource/%25t`.

May be better to have 3 letter service names externally? So we'd get:

- http://172.17.0.3:32733/pal/......
- ws://172.17.0.3:32733/frs/......
- http://172.17.0.3:32733/dta/.....

This would really differentiate ingress URI's from internal.

charts/filtered-resource-service/templates/ingress-route.yaml
dev-958: I can see that the problem is that I missed something when the service was developed. The path on the service should be just / so that routing becomes /resource/xxx-xxx-xxx-xxx, so make it resource/resource for now and raise a tiny ticket to correct the service.

dev930018: The issue here is the service has other endpoints (`/actuator/health`) so you'd have to not confuse the two:
`/actuator` the endpoint for health and stuff
vs
`/actuator` because for some reason the token for the request was "actuator"
~I'd change the `/resource` path in the service to be more like `/fetchResources` or similar~
See ticket created.

charts/filtered-resource-service/templates/pathparam-regex.yaml
dev930018: See ticket created

dev-958 (approved):

dev930018 (approved):

developer01189998819991197253 (approved):

PR #287, opened 2021-03-31
PAL-999Data, removed common import and replaced it with common code, …

developer01189998819991197253:
…updated imports and tests
As readers is already a dependency, adding readers common as a dependency is required

Moved token to common package


dbz120896 (dismissed):
Looks Good.

PR #289, opened 2021-04-01
Pal 1033 add synthetic data generator

pd104923:


Jenkinsfile
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@Library('jenkinsfile-lib')_
+@Library('jenkinsfile-lib@feature/PAL-1033-add-synth-data-gen')_
```
pd104923, line 16: Needs to be reverted before merge.

PR #316, opened 2021-05-21
Pal 1025 s3 k8s config

dev930018:


charts/filtered-resource-service/templates/statefulset.yaml
```diff
@@ -80,15 +80,9 @@ spec:
...
         volumeMounts:
-        - name: {{ include "filtered-resource-service.classpathJars.name" . }}
-          mountPath: {{ include "filtered-resource-service.classpathJars.mount" . }}
         - name: config-volume
           mountPath: {{ include "filtered-resource-service.config.path" . }}
       volumes:
-      - name: {{ include "filtered-resource-service.classpathJars.name" . }}
-        persistentVolumeClaim:
-          claimName: {{ include "filtered-resource-service.classpathJars.name" . }}
-          readOnly: true
```
pd104923, line 91: Is there a reason that these have been removed for the FRS?

pd104923 (changes requested):
One little question plus the license file should be reverted.

charts/filtered-resource-service/templates/statefulset.yaml
dev930018: Not needed, plus s3 akka-http conflicts with the version used in frs

pd104923 (dismissed):

developer01189998819991197253 (dismissed):

PR #329, opened 2021-06-15
Pal 1076 helm charts include released images

m78233:


user-service/pom.xml
```diff
@@ -278,6 +278,24 @@
...
+                            <goal>read-project-properties</goal>
+                        </goals>
+                        <configuration>
+                            <files>
+                                <file>${project.basedir}/../charts/${project.artifactId}/values.yaml</file>
+                            </files>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
```
pd104923: Is there a reason this has been duplicated? It's the same as the plugin above.

pd104923 (changes requested):
Just the 1 question

user-service/pom.xml
m78233: No, should not be there.  Have removed it.

pd104923 (dismissed):

pom.xml
```diff
@@ -43,6 +43,7 @@
     <properties>
         <!-- Project properties -->
         <java.version>11</java.version>
+        <version.number>0.5.0</version.number>
```
dev930018, line 46: I'd also expect the project-level `<version/>` tag to be updated appropriately for this parameter:
From
```
    <version>0.5.0-${revision}</version>
```
To
```
    <version>${version.number}-${revision}</version>
```

dev930018 (changes requested):

pom.xml
pd104923: Is this comment from @dev930018 no longer relevant? Looks like this should be updated.

pd104923 (commented):
1 small follow up on a previous comment.

pom.xml
m78233: Would have preferred this approach, but this hits up agains an issue. Both version and revision are reserved elements in Maven and can be used in the way that we have in the first example. The element version.number does not exists at the point the version element is being assigned a value and so any attempt to us it fails.

pd104923 (approved):

dev930018 (approved):

ac74475 (approved):

PR #333, opened 2021-06-25
Pal 1057 dark mode

developer01189998819991197253:
Test by changing github into dark mode and exploring the code base as this branch


pd104923 (dismissed):
The changes to the png files look good and they can be seen in dark mode for both GitHub and IntelliJ, I have therefore approved the PR.
However, when going through the images I have noticed some discrepancies in the wording when a service interacts with Kafka. This is out-of-scope of the ticket but would definitely help tidy things up.

dev930018 (changes requested):
The READMEs in the doc directories have broken links

dev930018 (approved):

PR #340, opened 2021-07-06
PAL-1089 updated cache ttl values, updated data service webmvc timeout

pd104923:


m78233 (changes requested):
Reads like there is one more part to this ticket.  There is the removal the performance.directory from the application-static.yaml.

m78233 (approved):
My bad, wrong repo

developer01189998819991197253 (approved):

